PR #360, opened 2026-04-17
fix(ui): enable feedback on questionnaire edit + cap dialog height

biagiodistefano:
## Summary

Two small UI fixes that surfaced while reviewing demo vs prod behavior.

### 1. Feedback questionnaire type disabled on the edit page

The creation page (`questionnaires/new/+page.svelte`) already supports creating Feedback questionnaires (enabled since #256). The edit page (`questionnaires/[id]/+page.svelte`) renders `QuestionnaireFormFields.svelte`, which was extracted as part of the #337 refactor but kept the old pre-#256 template — `disabled` + "Coming soon" badge, and `onValueChange` only accepted `'admission'`.

Result: users could create a Feedback questionnaire but not change an existing questionnaire's type to Feedback, and a freshly-created draft that redirects to the edit page showed a disabled option. Fix brings the edit selector in line with creation.

### 2. Direct-invitation modal expanding off-screen

`src/lib/components/ui/dialog/dialog-content.svelte` had no default max-height or overflow. Long dialogs (e.g. the Create Invitation dialog with many ticket tiers) grew past the viewport with no way to scroll to the submit button. `EventTokenModal` worked because it opted in with `max-h-[90vh] overflow-y-auto`.

Instead of patching each caller, fix the default. `cn()` uses `tailwind-merge`, so:
- `EventTokenModal`'s explicit same classes dedupe cleanly.
- Any future dialog can still override by passing its own `max-h-*` / `overflow-*` classes.

This implicitly fixes the three `Dialog.Content` uses in `InvitationListTab.svelte` (Create/Edit/Bulk Edit).

## Not included
- A corresponding backend fix for stale `OrganizationMember.status=BANNED` rows after blacklist entries are deleted — tracked separately on the backend side.

## Test plan
- [ ] Open an existing questionnaire's edit page: the type dropdown should allow selecting Feedback (no "Coming soon" pill).
- [ ] Create a new Feedback questionnaire → land on the edit page → type still shows Feedback and is editable.
- [ ] On an event with 10+ ticket tiers, open "Create Invitation" in the org admin invitations tab: the modal should cap at 90vh and scroll internally; the Send button must be reachable.
- [ ] Edit Invitation and Bulk Edit Invitation dialogs: same scrolling behavior.
- [ ] EventTokenModal (Create/Edit invitation link): unchanged behavior, still scrolls correctly.

## Release Notes

* **New Features**
  * Feedback questionnaire type is now available for selection.

* **Improvements**
  * Enhanced dialog content display with improved scrolling behavior for better content visibility.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "revel-frontend",
-	"version": "1.48.0",
+	"version": "1.48.1",
 	"private": true,
 	"type": "module",
 	"scripts": {
--- a/src/lib/components/invitations/InvitationListTab.svelte
+++ b/src/lib/components/invitations/InvitationListTab.svelte
@@ -566,7 +566,7 @@
 
 <!-- Create Invitation Dialog -->
 <Dialog.Root open={showCreateDialog} onOpenChange={(open) => (showCreateDialog = open)}>
-	<Dialog.Content class="sm:max-w-[600px]">
+	<Dialog.Content class="flex max-h-[90dvh] flex-col sm:max-w-[600px]">
 		<Dialog.Header>
 			<Dialog.Title>{m['eventInvitationsAdmin.createInvitations']()}</Dialog.Title>
 			<Dialog.Description>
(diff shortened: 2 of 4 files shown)


biagiodistefano:
### Code review

Found 1 issue:

1. `DialogPrimitive.Close` is an `absolute`-positioned child of the same container that now has `overflow-y-auto`, so on dialogs that exceed `90vh` the close button scrolls out of view along with the content — hurting dismissibility (revel-frontend/CLAUDE.md calls out "WCAG 2.1 AA Compliance is mandatory" and "Keyboard navigation: All interactive elements must be keyboard accessible"). Consider moving scrolling to an inner wrapper around `{@render children?.()}` so the close button stays pinned, or applying `sticky` to the close button.

https://github.com/letsrevel/revel-frontend/blob/1fd5e7b43b5f94f58fb8c05dc9119b39387093c8/src/lib/components/ui/dialog/dialog-content.svelte#L23-L38

PR #431, opened 2026-06-15
fix(events): RSVP buttons render on ticketed events from list views

### LINKED ISSUE #430: Admin events list: Tickets/Attendees button routes to wrong page → 400
## Problem

On a ticketed event reached from a list/summary surface (dashboard, org page, series), RSVP buttons can render where "Buy Tickets" should be. Clicking RSVP then 400s with `"You must get a ticket for this event."` The backend is behaving correctly — this is an FE gating bug.

## Code findings — root cause

- ⚠️ The admin event card is **not** at fault: `AdminEventCard.svelte:82-84` correctly routes "Tickets" to `/org/{slug}/admin/events/{id}/tickets`, gated `{#if event.requires_ticket}` (`:239-247`). (Correct since PR #337 — do not change it.)
- The only RSVP POST caller is `EventRSVP.svelte:59` (`eventpublicattendanceRsvpEvent`), reached via `EventActionSidebar.svelte`. The gate that should hide RSVP for ticketed events is `!event.requires_ticket` (`EventActionSidebar.svelte:494`, `EventRSVP.svelte:298`).
- But `requires_ticket` is non-null only on `EventDetailSchema`; it's **`boolean | null` on list/summary schemas** (`EventInListSchema`, `src/lib/api/types.gen.ts:1296`, etc.). List-built event objects (`dashboard/+page.svelte:167,200`; `org/+page.svelte:133`) can pass `requires_ticket == null`, so `!requiresTicket` is `true` → RSVP renders on a ticketed event → `POST /{event_id}/rsvp/{answer}` → backend 400 (`reason_code="requires_ticket"`, `next_step="purchase_ticket"`).
- Separate, related bug: `IneligibilityActionButton.svelte:127-238` has **no branch for `next_step === 'purchase_ticket'`** (or `'rsvp'`) — it falls through to `console.log` (`:237`), so the "Buy Tickets" CTA can be **inert**.
- The "400 when blacklisting" report (triage O4) is the **same class of bug, not blacklist** — blacklist actions never hit `/rsvp/`. O4 folds into this issue.

## Fix (FE-only, no backend change)

1. Fail closed on null: `EventRSVP.svelte:298` → `requiresTicket === false`; treat `requires_ticket == null` as ticketed in `EventActionSidebar.svelte:494`.
2. Add a `purchase_ticket` / `rsvp` branch to `IneligibilityActionButton.handleClick` that opens the ticket modal (`onGetTicketsClick` / `openTicketTierModal`, wired at detail `+page.svelte:734,874`) instead of `console.log`.

Reported by @bechtold.

---
_From the Oskar feedback session (2026-06-14) — triage IDs **B1 / O4**._

biagiodistefano:
## Root cause

On a ticketed event reached from a **list/summary surface** (dashboard, org page, series), the event object is built from a list schema (`EventInListSchema`, etc.) where `requires_ticket` is typed `boolean | null` — unlike `EventDetailSchema`, where it is non-null `boolean`.

The RSVP gate used `!requires_ticket` (`EventRSVP.svelte`, `EventActionSidebar.svelte`). When `requires_ticket` is `null`, `!null` is `true`, so RSVP buttons rendered on a **ticketed** event. Clicking RSVP then hit `POST /{event_id}/rsvp/{answer}` and returned `400 — "You must get a ticket for this event."` The backend is correct; this was an FE gating bug caused by the nullable field.

Separately, `IneligibilityActionButton.handleClick` had **no branch** for `next_step === 'purchase_ticket'` (or `'rsvp'`) and fell through to `console.log`, so the "Buy Tickets" CTA could be inert.

## Fix (FE-only)

1. **Fail closed on null.** Treat unknown/null `requires_ticket` as **ticketed** (hide RSVP):
   - `EventRSVP.svelte`: gate on `requiresTicket === false`; prop widened to `boolean | null` so the null case is preserved (not coerced upstream).
   - `EventActionSidebar.svelte`: RSVP branch gated on `event.requires_ticket === false`.
2. **Make the Buy CTA work.** Added a `purchase_ticket` branch to `IneligibilityActionButton.handleClick` that opens the ticket-tier purchase modal via a threaded `onGetTicketsClick` callback (`EventActionSidebar → EligibilityStatusDisplay → IneligibilityActionButton`), plus an `rsvp` branch that routes to the event page. The existing `onGetTicketsClick` wiring (`openTicketTierModal` on the detail page) is reused.

`AdminEventCard.svelte` was not touched (already correct since #337).

## Manual verification

- From `/dashboard` (or an org page / series), open a **ticketed** event you have **no** ticket for: you should see **Buy Tickets**, never RSVP.
- Clicking **Buy Tickets** opens the ticket-tier modal — no 400.
- A **free** (non-ticketed) event still shows RSVP buttons and RSVP succeeds.

Closes #430

## Summary by CodeRabbit

* **New Features**
  * Integrated direct ticket purchase action into the event eligibility status display for seamless ticket acquisition.

* **Bug Fixes**
  * Enhanced RSVP flow logic to correctly show RSVP options only for explicitly non-ticketed events.
  * Improved event eligibility flows to properly handle events with unknown ticket requirements.

### DIFF
--- a/src/lib/components/events/EligibilityStatusDisplay.svelte
+++ b/src/lib/components/events/EligibilityStatusDisplay.svelte
@@ -36,6 +36,8 @@
 		applyBefore?: string | null;
 		onInvitationRequestSuccess?: () => void;
 		onWhitelistRequestSuccess?: () => void;
+		/** Opens the ticket-tier purchase modal (for the `purchase_ticket` step). */
+		onGetTicketsClick?: () => void;
 		class?: string;
 	}
 
@@ -50,6 +52,7 @@
 		applyBefore,
 		onInvitationRequestSuccess,
 		onWhitelistRequestSuccess,
+		onGetTicketsClick,
 		class: className
 	}: Props = $props();
 
@@ -278,6 +281,7 @@
 						retryOn={eligibility.retry_on}
 						{onInvitationRequestSuccess}
 						{onWhitelistRequestSuccess}
+						{onGetTicketsClick}
 					/>
 				</div>
 			{/if}
--- a/src/lib/components/events/EventActionSidebar.svelte
+++ b/src/lib/components/events/EventActionSidebar.svelte
@@ -490,8 +490,13 @@
 
 		<!-- Primary Action (if not attending) -->
 		{#if !isAttending}
-			<!-- RSVP Flow for non-ticketed events -->
-			{#if !event.requires_ticket}
+			<!-- RSVP Flow for non-ticketed events.
+			     Fail closed: only show RSVP when the event is explicitly known to
+			     be non-ticketed. `requires_ticket` is `boolean | null` on
+			     list/summary-built event objects, where null means unknown; a null
+			     value is treated as ticketed so RSVP is never shown on a ticketed
+			     event reached from a list surface (issue #430). -->
+			{#if event.requires_ticket === false}
 				<EventRSVP
 					eventId={event.id}
 					eventName={event.name}
@@ -521,6 +526,7 @@
 							applyBefore={event.apply_before}
 							{onInvitationRequestSuccess}
 							{onWhitelistRequestSuccess}
+							{onGetTicketsClick}
 						/>
 					</div>
 				{:else}
--- a/src/lib/components/events/EventRSVP.svelte
+++ b/src/lib/components/events/EventRSVP.svelte
@@ -19,7 +19,12 @@
 		eventName: string;
 		userStatus: UserEventStatus | null;
 		isAuthenticated: boolean;
-		requiresTicket: boolean;
+		/**
+		 * Whether the event requires a ticket. Non-null on EventDetailSchema but
+		 * `boolean | null` on list/summary schemas, where `null` means unknown.
+		 * The RSVP gate fails closed on null (treats it as ticketed).
+		 */
+		requiresTicket: boolean | null;
 		event?: EventDetailSchema;
 		eventTokenDetails?: EventTokenSchema | null;
 		onGuestRsvpClick?: () => void;
@@ -292,10 +297,12 @@
 		return eligibilityStatus && !eligibilityStatus.allowed && !showSuccess;
 	});
 
-	// Show RSVP if:
-	// 1. Event doesn't require tickets
-	// 2. Always show for non-ticketed events (to display login prompt if needed)
-	const shouldRender = $derived(!requiresTicket);
+	// Show RSVP only when the event is explicitly known to be non-ticketed.
+	// `requiresTicket` is non-null on EventDetailSchema, but list/summary
+	// schemas type it as `boolean | null`. A null value means "unknown", so we
+	// fail closed (treat as ticketed) to avoid rendering RSVP on a ticketed
+	// event reached from a list surface, which would 400 on submit (issue #430).
+	const shouldRender = $derived(requiresTicket === false);
 
 	// Check if we should show guest button instead of regular RSVP
 	const shouldShowGuestButton = $derived(
--- a/src/lib/components/events/IneligibilityActionButton.svelte
+++ b/src/lib/components/events/IneligibilityActionButton.svelte
@@ -41,6 +41,8 @@
 		eventTokenDetails?: EventTokenSchema | null;
 		onInvitationRequestSuccess?: () => void;
 		onWhitelistRequestSuccess?: () => void;
+		/** Opens the ticket-tier purchase modal (for the `purchase_ticket` step). */
+		onGetTicketsClick?: () => void;
 		class?: string;
 	}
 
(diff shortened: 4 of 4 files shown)


biagiodistefano:
## Code Review — PR #431 (`fix/430-rsvp-ticketed-gating`)

Reviewed `git diff 141b3480..a15ed988` (4 files) with surrounding context, the consumer chain (detail `+page.svelte`), the generated schema types, and the sibling `IneligibilityMessage` path. Verified against issue #430 and the three requirements.

### Strengths

- **Null-safety is correct and complete.** Both gates now fail closed: `EventRSVP.svelte:305` (`requiresTicket === false`) and `EventActionSidebar.svelte:499` (`event.requires_ticket === false`). No `!requires_ticket` / `!requiresTicket` gating remains in any of the touched components (the only other `requiresTicket={false}` hit is inside a JSDoc example comment, `EventRSVP.svelte:337`).
- **Distinction preserved, not coerced.** `requiresTicket` is widened to `boolean | null` at the `EventRSVP` prop (`:27`) rather than forcing `null → false` upstream, exactly as required. `EventActionSidebar` passes `event.requires_ticket` (non-null `EventDetailSchema.requires_ticket: boolean`) into the now-`boolean | null` prop — valid widening, so `make types` is unaffected. No `any`, no upstream coercion.
- **Callback threading is sound.** `onGetTicketsClick` flows detail `+page.svelte` (`onGetTicketsClick={openTicketTierModal}`, both sidebar instances at :734 and :874) → `EventActionSidebar` → `EligibilityStatusDisplay:284` → `IneligibilityActionButton`, and the new `purchase_ticket` branch calls `onGetTicketsClick?.()` which is `openTicketTierModal`. The Buy CTA now opens the tier modal instead of being inert.
- **Regression-safe.** Free events (`requires_ticket === false`) still render RSVP. Invitation/whitelist/waitlist `next_step` branches are untouched and still return before the new branches. The secondary "change RSVP" `EventRSVP` (sidebar :643) and the ticketed `ActionButton` (:536) are unaffected.
- **Requirement #3 met:** `AdminEventCard.svelte` has zero diff in this range.
- Svelte 5 runes usage is correct (`$derived` gate, `$props()`), comments clearly explain the fail-closed rationale, and a11y of the changed `<Button>` is unchanged (still keyboard/screen-reader accessible).

### Issues

**Minor — `IneligibilityMessage.svelte` renders the same button without `onGetTicketsClick` (latent inert Buy CTA).**
`IneligibilityMessage.svelte:482` renders `IneligibilityActionButton` but does **not** thread `onGetTicketsClick`. If a `purchase_ticket` next_step ever reaches that surface, the new branch calls `onGetTicketsClick?.()` → `undefined` → no-op (silently inert, the exact class of bug #430 set out to kill). In practice this path is reached via `EventRSVP`, which now only renders when `requiresTicket === false` (non-ticketed), so `purchase_ticket` shouldn't surface here today — hence Minor, not blocking. Worth either threading the callback through `IneligibilityMessage` for symmetry/future-proofing, or adding a short comment noting why it's intentionally omitted.

**Minor — `rsvp` branch reachability / hard navigation.**
`IneligibilityActionButton.handleClick` `rsvp` branch (new) uses `window.location.href` to route to the event page. (a) This button only renders under `!eligibility.allowed` (`EligibilityStatusDisplay:270`), so an `rsvp` next_step here is an edge case; the branch is defensive and fine. (b) Style nit: the rest of the file already uses `window.location.href`, so this is consistent — but a full-page reload (vs. `goto`) is heavier than needed. Non-blocking; matches existing file conventions.

**Minor — pre-existing catch-all `console.log` remains (`IneligibilityActionButton.svelte:255`).**
The `console.log('Action clicked', ...)` fallthrough still exists, but it is pre-existing (not introduced here) and is now unreachable for `purchase_ticket`/`rsvp` since both return earlier. The PR's "no dead console.log" obligation applied to the two new branches, which is satisfied. Cleaning up the leftover catch-all log would be a nice unrelated tidy but is out of scope.

### Recommendations

1. (Optional, low effort) Thread `onGetTicketsClick` through `IneligibilityMessage.svelte → IneligibilityActionButton` for parity with `EligibilityStatusDisplay`, closing the latent inert-CTA path before it can regress.
2. (Optional) Consider `goto()` over `window.location.href` in the new `rsvp` branch for SPA navigation — only if you want to diverge from the file's existing pattern.
3. No change needed for the gating logic or type widening — both are correct as written.

### Assessment

**Ready to merge: Yes.** The core fix is correct, type-safe, and minimal: both RSVP gates fail closed on `null`, the nullable distinction is preserved end-to-end, the Buy CTA is now wired through to the tier modal, free-event RSVP and other eligibility states are unregressed, and `AdminEventCard` is untouched. The only findings are Minor/latent (an unthreaded sibling button and a pre-existing catch-all log) that don't affect the shipped behavior for issue #430; address recommendation #1 in this PR or a fast follow-up at your discretion.

PR #433, opened 2026-06-15
fix: small UX quick wins from Oskar feedback (Q3, D3, CI3)

### LINKED ISSUE #421: fix(ux): co-locate Save button with 'Cancel Editing' in questionnaire edit mode
## Problem

In questionnaire edit mode, the Save action is hard to find. The "Cancel Editing" control sits at the top in the edit-mode banner, but Save is a full scroll away at the page bottom — so users can't see how to save.

## Code findings

- Save (`saveQuestionnaire`, `src/routes/(auth)/org/[slug]/admin/questionnaires/[id]/+page.svelte:318`) is triggered by a button in a bottom "Actions" row (`:810-818`).
- "Cancel Editing" lives at the top in the green edit-mode banner (`:438-440`), which only flips `isEditMode`. The banner even says "Changes will be saved when you click Save" (`:433-434`) — but Save is nowhere near it.

## Fix

Add a primary **Save** button into the edit-mode banner row (`:427-441`), wired to the existing `saveQuestionnaire`. Optionally make the banner sticky so Save/Cancel stay reachable while scrolling. No backend change.

---
_From the Oskar feedback session (2026-06-14) — triage ID **Q3**._

### LINKED ISSUE #425: fix(ux): show expiration date instead of countdown in invite-link list
## Problem

The invitation-link list shows a relative **countdown** for expiry ("3 days") instead of the actual expiration **date**, so organizers have to re-open each link to see when it really expires.

## Code findings

- `src/lib/components/tokens/EventTokenCard.svelte:28` derives `expirationDisplay = getExpirationDisplay(token.expires_at)` (rendered in the column at `:74-76`).
- The helper `getExpirationDisplay` (`src/lib/utils/tokens.ts:79-102`) returns a relative countdown ("3 days", "Less than 1 hour", "Never", "Expired") — never the date.

## Fix

Render the absolute date via the existing `formatDateTime` (`src/lib/utils/date.ts:235`), keeping the "Never"/null case; optionally keep the countdown as a hover tooltip. `getExpirationDisplay` is only used here, so it's safe to replace.

> Related: pairs naturally with the invite-link "duration → expiration date" copy fix (triage D2) if that lands together.

---
_From the Oskar feedback session (2026-06-14) — triage ID **D3**._

### LINKED ISSUE #428: fix(ux): add clear button to check-in attendee search field
## Problem

The attendee-search field used during check-in has no clear/X button, which is slow on mobile when doing repeated searches.

## Code findings

- Both admin search fields are bare `<Input type="search">` with a leading icon and no trailing clear button:
  - `src/routes/(auth)/org/[slug]/admin/events/[event_id]/attendees/+page.svelte:510-522`
  - `src/lib/components/tickets/TicketFilters.svelte:27-40`
- A reusable solution already exists: `src/lib/components/events/filters/SearchInput.svelte` has a built-in clear button (`:69-78`, `aria-label` key `common.search_clear` at `messages/en.json:687`).

## Fix

Reuse `SearchInput.svelte` in both pages (promote it to `common/` and make its currently-hardcoded aria-label a prop), or add the same `{#if value}`-gated clear button inline.

---
_From the Oskar feedback session (2026-06-14) — triage ID **CI3**._

biagiodistefano:
Three FE-only UX quick wins from Oskar's feedback.

## Changes

**Q3 — Save next to 'Cancel Editing' (questionnaire editor)**
Adds a primary Save button into the edit-mode banner row, next to "Cancel Editing", wired to `saveQuestionnaire` with the existing `isSaving`/`canEdit` state. Editors no longer need to scroll to the bottom Actions row to save. The bottom Save button is kept as a secondary affordance.

**D3 — Absolute expiration date in invite-link list**
`EventTokenCard` now renders the absolute expiration date via `formatDateTime` instead of a relative countdown. New `getExpirationDate` helper preserves the "Never" case for null expiry; the countdown is kept as a `title` tooltip.

**CI3 — Clear button on check-in / attendee search**
Reuses `SearchInput` (which has a built-in, translatable clear button) in the attendees page and the ticket filters, replacing bare `Input` fields that had no way to clear the query. `SearchInput`'s aria-label is now a prop so each usage stays accessible and translatable.

## Verification (manual)
- Questionnaire editor: enter edit mode, confirm Save appears in the green banner, respects disabled state while saving, and saves correctly; bottom Save still works.
- Invite-link list: confirm cards show an absolute date (and "Never" for tokens with no expiry); hover shows the countdown tooltip.
- Attendees + ticket filters: type a query, confirm the X clear button appears, clears the field on click, is keyboard-focusable with an accessible name, and re-applies filters.

Closes #421
Closes #425
Closes #428

### DIFF
--- a/src/lib/components/events/filters/SearchInput.svelte
+++ b/src/lib/components/events/filters/SearchInput.svelte
@@ -7,6 +7,7 @@
 		value: string;
 		onSearch: (value: string) => void;
 		placeholder?: string;
+		ariaLabel?: string;
 		debounceMs?: number;
 		class?: string;
 	}
@@ -15,6 +16,7 @@
 		value = '',
 		onSearch,
 		placeholder = m['filters.search.eventsPlaceholder'](),
+		ariaLabel = m['filters.search.eventsLabel'](),
 		debounceMs = 300,
 		class: className
 	}: Props = $props();
@@ -64,7 +66,7 @@
 			oninput={handleInput}
 			{placeholder}
 			class="h-10 w-full rounded-md border border-input bg-background pl-9 pr-9 text-sm ring-offset-background transition-colors placeholder:text-muted-foreground focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 disabled:cursor-not-allowed disabled:opacity-50"
-			aria-label={m['filters.search.eventsLabel']()}
+			aria-label={ariaLabel}
 		/>
 		{#if inputValue}
 			<button
--- a/src/lib/components/tickets/TicketFilters.svelte
+++ b/src/lib/components/tickets/TicketFilters.svelte
@@ -1,13 +1,12 @@
 <script lang="ts">
 	import * as m from '$lib/paraglide/messages.js';
-	import { Search } from 'lucide-svelte';
-	import { Input } from '$lib/components/ui/input';
+	import SearchInput from '$lib/components/events/filters/SearchInput.svelte';
 
 	interface Props {
 		searchQuery: string;
 		selectedStatus: string | null;
 		selectedPaymentMethod: string | null;
-		onSearch: (e: Event) => void;
+		onSearch: (value: string) => void;
 		onStatusFilter: (status: string | null) => void;
 		onPaymentMethodFilter: (method: string | null) => void;
 	}
@@ -24,20 +23,12 @@
 
 <div class="mt-6 space-y-4">
 	<!-- Search -->
-	<div class="relative">
-		<Search
-			class="absolute left-3 top-1/2 h-4 w-4 -translate-y-1/2 text-muted-foreground"
-			aria-hidden="true"
-		/>
-		<Input
-			type="search"
-			placeholder={m['eventTicketsAdmin.searchPlaceholder']()}
-			value={searchQuery}
-			oninput={onSearch}
-			class="pl-10"
-			aria-label="Search tickets"
-		/>
-	</div>
+	<SearchInput
+		value={searchQuery}
+		{onSearch}
+		placeholder={m['eventTicketsAdmin.searchPlaceholder']()}
+		ariaLabel={m['eventTicketsAdmin.searchPlaceholder']()}
+	/>
 
 	<!-- Status Filters -->
 	<div>
--- a/src/lib/components/tokens/EventTokenCard.svelte
+++ b/src/lib/components/tokens/EventTokenCard.svelte
@@ -7,6 +7,7 @@
 	import {
 		getEventTokenStatus,
 		formatTokenUsage,
+		getExpirationDate,
 		getExpirationDisplay,
 		getEventTokenUrl
 	} from '$lib/utils/tokens';
@@ -25,7 +26,8 @@
 
 	const status = $derived(getEventTokenStatus(token));
 	const usageDisplay = $derived(formatTokenUsage(token.uses, token.max_uses));
-	const expirationDisplay = $derived(getExpirationDisplay(token.expires_at));
+	const expirationDate = $derived(getExpirationDate(token.expires_at));
+	const expirationCountdown = $derived(getExpirationDisplay(token.expires_at));
 
 	async function copyLink() {
 		if (!token.id) {
@@ -72,7 +74,7 @@
 				</div>
 				<div>
 					<span class="font-medium">{m['eventTokenCard.expires']()}</span>
-					{expirationDisplay}
+					<span title={expirationCountdown}>{expirationDate}</span>
 				</div>
 			</div>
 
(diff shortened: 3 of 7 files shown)


biagiodistefano:
# Code Review — PR #433 (`fix/oskar-ux-quickwins`)

Three isolated UX quick-wins: Q3 (questionnaire Save button), D3 (token absolute expiration date), CI3 (reuse `SearchInput`). Reviewed `git diff 141b3480..5c55be2d` (7 files) plus full `SearchInput.svelte` and both consuming pages at HEAD.

## Strengths

- **CI3 is a genuine simplification, not just a swap.** Both parents drop their hand-rolled `setTimeout` debounce + `(e: Event)` target-casting in favor of `SearchInput`'s internal debounce and clean `(value: string)` callback. Less code, one debounce source of truth, and the clear button + `type="search"` come for free. The `searchQuery`/`searchInput` state and `applyFilters()` URL wiring are preserved, so the value that reaches the query is unchanged.
- **No double-debounce.** The old timers were fully removed from both pages; `handleSearch` now applies filters synchronously and relies solely on `SearchInput`'s 300ms debounce. Correct.
- **Initial value from URL params is preserved.** `value={searchInput}` / `value={searchQuery}` (both seeded from `data.filters.search`) flow into `SearchInput`, whose `$effect` syncs `inputValue = value`. The pre-filled search term survives a reload.
- **`ariaLabel` prop is backwards-compatible** (defaults to the existing `filters.search.eventsLabel`), so the events filter callers are unaffected. Both new callers pass an explicit label.
- **Clear button a11y is solid** (pre-existing in `SearchInput`): real `<button type="button">`, `aria-label={m['common.search_clear']()}`, visible `focus-visible` ring, keyboard-operable. Inherited correctly.
- **D3 / Q3 reuse existing, locale-aware helpers** (`formatDateTime`, message keys for Save/Saving) rather than inlining — consistent with the codebase.
- **Q3 Save button** is wired to the same `saveQuestionnaire` and the same `disabled={isSaving || !canEdit}` guard as the bottom button, and `saveQuestionnaireIncremental` is awaited under the `isSaving` flag — no duplicate-submit risk even with two buttons.

## Issues

### Critical
None.

### Important
None blocking. See Minor items — all are low-impact and consistent with existing baselines.

### Minor

1. **`tokens.ts:108` (D3) — hardcoded `'Never'` string is not localized.**
   `getExpirationDate` returns the literal `'Never'`, mirroring the pre-existing `getExpirationDisplay` which also returns `'Never'`/`'Expired'`. So this is *consistent* with current code, but it's a user-facing English string in an otherwise i18n'd app. Not introduced by this PR's intent and not worth blocking, but worth a follow-up issue to route both helpers through `m[...]`. Fix (future): `return m['eventTokenCard.never']();`.

2. **CI3 — attendees debounce timing changed 500ms → 300ms.**
   The old attendees handler debounced at 500ms; `SearchInput` defaults to 300ms (tickets was already 300ms, so unchanged there). Functionally fine and arguably better/consistent, but it *is* a behavior change. If 500ms was intentional (heavier attendee list), pass `debounceMs={500}`. Otherwise leave as-is — 300ms is the better default. Flagging for awareness only.

3. **D3 tooltip is mouse/focus-only (`title` attribute).**
   Moving the countdown into `title={expirationCountdown}` on a `<span>` means the countdown is invisible to touch users and not reliably exposed to screen readers (`title` on a non-interactive `<span>` has inconsistent AT support). The absolute date is now the primary, always-visible value (good), and the countdown is genuinely secondary, so this is acceptable for a quick-win. If the countdown is considered important info, prefer a real tooltip component or visible secondary text. Acceptable as scoped.

4. **CI3 — redundant local `searchInput` state in attendees.**
   `handleSearch` sets both `searchInput = value` and `searchQuery = value`, but `SearchInput` already owns its display state internally; the parent's `searchInput` is now only used to feed `value=` back in (a no-op round-trip since they're always equal) and for the seed. Could be collapsed to a single `searchQuery` state like the tickets page does, for symmetry. Pure cleanup, not required.

## Recommendations

- Open a follow-up to localize `'Never'`/`'Expired'` in both `getExpirationDate` and `getExpirationDisplay` (Minor #1).
- Confirm the attendees 300ms vs prior 500ms is intended (Minor #2); pass `debounceMs={500}` if not.
- Optional: collapse attendees' `searchInput`+`searchQuery` to one state (Minor #4).
- Quick manual smoke before merge: (a) load attendees/tickets with `?search=foo` and confirm the box is pre-filled; (b) type, confirm one debounced URL update; (c) click the X clear button and confirm it resets to the unfiltered list.

## Assessment

**Ready to merge: Yes.** All three fixes are correctly wired, Svelte 5 runes / TS-strict clean, and the highest-risk change (CI3) preserves search semantics — initial URL value, single 300ms debounce, the value reaching the query, and an accessible clear button. No handler signature mismatches: both parents updated to `(value: string)` matching `SearchInput`'s `onSearch`. The only deltas are minor, non-blocking, and mostly pre-existing patterns (un-localized "Never", `title` tooltip). Recommend merging; address Minor #1/#2 as small follow-ups.

PR #467, opened 2026-06-17
fix(theme): eliminate dark-mode flash (FOUC) on SSR loads (#440)

### LINKED ISSUE #440: fix(theme): eliminate dark-mode flash (FOUC) with anti-FOUC inline script
## Context

From the Oskar feedback session (2026-06-14): a light-mode flash (FOUC) appears on navigation/initial load for users who have selected dark mode.

## Code findings (verified against `main`)

- Theme is managed by `mode-watcher` (`<ModeWatcher />` at `src/routes/+layout.svelte:282`; `ThemeToggle.svelte` → `setMode`). The selected mode lives in `localStorage` (`mode-watcher-mode`) and is only applied **after hydration**.
- There is **no anti-FOUC inline `<head>` script**: `src/app.html` has no theme logic, and `src/hooks.server.ts` does zero theme handling (`+layout.server.ts` handles only auth). SSR is on globally, so the server renders light → the browser paints light → the client flips to dark = the visible flash.

This is architectural (mode-watcher applies the class post-hydration), not a regression — but it's a real, fixable polish issue.

## Proposed fix

Add a **synchronous** (no `defer`/`async`) inline script to `src/app.html` `<head>`, before `%sveltekit.head%`, that reads `localStorage.getItem('mode-watcher-mode')` (falling back to `prefers-color-scheme`) and sets `document.documentElement.classList` / `style.colorScheme` before first paint — the standard mode-watcher anti-FOUC pattern.

**Optional hardening:** also persist the mode to a cookie and emit `class="dark"` on `<html>` from `hooks.server.ts`, killing the flash even before JS runs.

> Verify the exact `mode-watcher` storage key for the installed version before wiring.

## Acceptance

- A dark-mode user sees no light flash on hard load or navigation.
- Light-mode and system-preference users are unaffected.

_Source: Oskar session, sparse notes L11, transcript ~00:08._

biagiodistefano:
## Summary

Dark-mode users saw a **light flash (FOUC)** on initial load and navigation. Closes #440.

### Root cause (production-only)

`<ModeWatcher />` *does* inject an anti-FOUC snippet via `<svelte:head>` using `setInitialMode.toString()` — but it's emitted through `{@html}` **without a nonce**. Under our production CSP (`script-src 'self'`, and SvelteKit's `mode: 'auto'` resolves to **nonce-based** for SSR pages), SvelteKit only nonces the scripts *it* generates; an arbitrary `{@html}` script is not nonced and is therefore **blocked**. So the snippet never ran: the server painted light → the client flipped to dark after hydration → the visible flash.

This is why dev (CSP off) looked fine while demo/prod flashed.

### Fix

Add a **synchronous, nonced** inline script to `src/app.html` `<head>` (before `%sveltekit.head%`) that applies the `dark` class and `color-scheme` before first paint. It carries `nonce="%sveltekit.nonce%"`, which SvelteKit substitutes per-request so it passes CSP. It mirrors mode-watcher exactly:

- storage key `mode-watcher-mode`, values `dark` / `light` / `system` (defaulting to `system`)
- dark class `dark`
- `system` falls back to `prefers-color-scheme`

mode-watcher's `onMount` still drives runtime toggling and keeps everything in sync after hydration.

> Verified there is **no prerendering** in the app, so introducing `%sveltekit.nonce%` (which is incompatible with prerendering) is safe.

### Acceptance

- ✅ Dark-mode user sees no light flash on hard load or navigation.
- ✅ Light-mode and system-preference users are unaffected.

## Testing

New E2E spec `tests/e2e/dark-mode-fouc.spec.ts`, which runs against the **production build+preview** (the only place CSP is enforced):

1. The nonced anti-FOUC script is present, synchronous (no `defer`/`async`), and its nonce **is allow-listed** by the `content-security-policy` header.
2. A dark-mode user (seeded `localStorage`) lands with the `dark` class and `color-scheme: dark`.
3. An explicit light-mode user is **not** forced to dark.

Verified locally against a production preview — all 3 pass; manual `curl` confirms `script-src 'self' 'nonce-…'` matches the script's nonce. `make types` clean (0 errors); lint/format clean on the changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved flash of unstyled content when dark mode is active. Dark mode preference now applies smoothly during page load, respecting both user settings and system preferences.

* **Tests**
  * Added regression tests to verify dark mode functionality works correctly in production.

### DIFF
--- a/src/app.html
+++ b/src/app.html
@@ -24,6 +24,29 @@
 		<meta property="og:image:type" content="image/png" />
 		<meta name="twitter:card" content="summary_large_image" />
 		<meta name="twitter:image" content="%sveltekit.assets%/og-image.png" />
+		<!--
+			Anti-FOUC: apply the dark class before first paint so dark-mode users
+			never see a light flash on SSR loads (#440). mode-watcher injects an
+			equivalent <svelte:head> snippet, but it carries no nonce and is blocked
+			by our production CSP (script-src 'self'); this script is nonced via
+			%sveltekit.nonce% so it runs. Must stay synchronous (no defer/async) and
+			mirror mode-watcher's storage key ('mode-watcher-mode') and dark class.
+		-->
+		<script nonce="%sveltekit.nonce%">
+			(() => {
+				try {
+					const mode = localStorage.getItem('mode-watcher-mode') || 'system';
+					const dark =
+						mode === 'dark' ||
+						(mode === 'system' && window.matchMedia('(prefers-color-scheme: dark)').matches);
+					const root = document.documentElement;
+					root.classList.toggle('dark', dark);
+					root.style.colorScheme = dark ? 'dark' : 'light';
+				} catch {
+					/* localStorage unavailable (e.g. private mode) — fall back to light */
+				}
+			})();
+		</script>
 		%sveltekit.head%
 	</head>
 	<body data-sveltekit-preload-data="hover">
--- a/tests/e2e/dark-mode-fouc.spec.ts
+++ b/tests/e2e/dark-mode-fouc.spec.ts
@@ -1,0 +1,62 @@
+import { test, expect } from '@playwright/test';
+
+// Regression test for #440: dark-mode users saw a light flash (FOUC) on SSR
+// loads. The fix is a synchronous, nonced inline script in app.html that applies
+// the `dark` class before first paint. mode-watcher injects an equivalent
+// snippet, but it carries no nonce and is blocked by the production CSP
+// (script-src 'self'), so it never ran — hence the flash.
+//
+// These tests run against the production build+preview server (see
+// playwright.config.ts), where CSP is enforced — the only place the bug
+// reproduces. They would fail on `main` (no nonced anti-FOUC script).
+
+test.describe('dark-mode anti-FOUC (#440)', () => {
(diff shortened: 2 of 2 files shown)


biagiodistefano:
Thanks @coderabbitai 🙏 — addressed in e7828ed:

- **Extract localStorage setup into a typed helper** ✅ Done. Replaced the two inline `addInitScript` calls with `seedModePreference(page, mode)`, matching the `loginAsAliceOwner(page)` convention in `polls-admin.spec.ts`.

- **"Assertions can miss first-paint flash" / add a client-side nav step** — Respectfully skipping, with reasoning:
  - The anti-FOUC `<script>` is **synchronous in `<head>`**, so the `dark` class is applied before the body paints by definition; there is no intermediate state for Playwright to observe. The first test already guards the real mechanism — it asserts the script is present, **not** `defer`/`async`, and that its nonce is allow-listed by the CSP header (this is exactly what was broken: the un-nonced mode-watcher snippet was CSP-blocked in prod). That test fails on `main`.
  - FOUC only occurs on **full document (SSR) loads**. Client-side SPA navigation reuses the live DOM and never re-paints `<html>`, so a nav step wouldn't exercise the flash path.

Happy to revisit if you see a gap in that reasoning.

biagiodistefano:
### Code review

PR #486, opened 2026-06-21
fix(auth): reflect login in navbar without a manual refresh (#485)

### LINKED ISSUE #485: Navbar shows logged-out state after login until a manual refresh
## Summary

After logging in, the navbar continues to render the **logged-out** state (Login / Sign up buttons, public nav) until the user manually refreshes the page. The in-memory auth store is never bootstrapped on the client following the post-login redirect, so the UI is out of sync with the actual (authenticated) session.

This is especially visible on **demo.letsrevel.io**: the demo DB resets every 24h, so a returning user arrives with a stale cookie and a "semi-broken User" in the navbar. They log out to start fresh, then log back in — and the navbar *still* shows Login until they refresh. Confusing and looks broken.

## Steps to reproduce

1. Log in with valid credentials (email/password, or a demo account).
2. You are redirected to `/dashboard`.
3. **Observe:** the navbar still shows **Login / Sign up** (logged-out chrome), even though you're authenticated and on the dashboard.
4. Manually refresh the page → navbar now correctly shows the user menu.

Variant (the demo flow): start with a stale session from a previous day → click Logout → click Login → after login the navbar shows Login until refresh.

## Expected vs actual

- **Expected:** immediately after a successful login + redirect, the navbar reflects the authenticated state (user menu, notifications, admin button) with no manual refresh.
- **Actual:** navbar stays logged-out until a full page reload.

## Root cause

Login and logout are **asymmetric** in how they update client-side auth state:

- **Logout** (`src/lib/components/common/UserMenu.svelte` → `handleLogout`) clears the store synchronously (`authStore.logout()` + `queryClient.clear()`) *and* navigates to `/logout`, whose `+page.server.ts` clears cookies and does a **full-page** `throw redirect(303, '/?logged_out=true')`. The navbar updates instantly and SSR re-runs. ✅
- **Login** (`src/routes/(public)/login/+page.server.ts`, `?/login` action) only sets the `access_token` / `refresh_token` cookies and `throw redirect(303, '/dashboard')`. It never touches the in-memory `authStore`. The form uses `use:enhance` with `await update()`, so the redirect is followed as a **client-side** navigation (SPA), not a full reload.

The navbar's authenticated state is driven by `authStore.isAuthenticated` (`Header.svelte`), which is bootstrapped by the `$effect` in `src/routes/+layout.svelte`. That effect only fires when `data.auth.hasAccessToken` flips `false → true`. `data.auth.hasAccessToken` is produced by `src/routes/+layout.server.ts`, whose `load` reads only `cookies`/`locals` and declares **no `url` / `params` / `depends()` dependency**.

During the client-side redirect navigation that follows login, SvelteKit does **not** re-run that layout `load` (nothing it tracks changed, and a redirect result doesn't `invalidateAll`). So `data.auth.hasAccessToken` stays `false` (its value from when `/login` was rendered), the bootstrap `$effect` never runs, and the navbar stays logged-out. A manual refresh triggers a fresh SSR load where the cookie is present → `hasAccessToken` is `true` → bootstrap runs → navbar correct.

### Relevant files
- `src/routes/(public)/login/+page.server.ts` — login action: sets cookies, `redirect(303)`, no store/invalidation.
- `src/routes/(public)/login/+page.svelte` — `use:enhance` callback (`await update()`).
- `src/routes/+layout.server.ts` — `auth.hasAccessToken` flag; no `depends('app:auth')`.
- `src/routes/+layout.svelte` — `$effect` that bootstraps `authStore` from `data.auth.hasAccessToken`.
- `src/lib/components/common/Header.svelte` — navbar reads `authStore.isAuthenticated`.
- `src/lib/components/common/UserMenu.svelte` — logout (the working, symmetric path) for comparison.

## Suggested fixes (pick one — discuss before implementing)

1. **Force a layout re-bootstrap after login (preferred, keeps SPA nav).** Add `depends('app:auth')` to `+layout.server.ts`, and in the login page's `enhance` callback call `await invalidateAll()` (or `invalidate('app:auth')`) after the redirect lands so the auth flag is recomputed and the `$effect` fires.
2. **Make login symmetric with logout (simplest, robust).** Do a full-page navigation to the destination after a successful login (e.g. `window.location.assign(returnUrl)`), so SSR re-runs with the new cookie — mirrors the logout path. Costs the SPA transition.
3. **Bootstrap the store client-side on login**, the mirror image of `handleLogout`: after the action succeeds, populate `authStore` (via `/api/auth/refresh`) before/at navigation, instead of waiting on the layout flag.

## Acceptance criteria
- After a successful login (password and 2FA paths) the navbar reflects the authenticated state with **no manual refresh**.
- The logout → login round trip on the demo no longer shows a stale logged-out navbar.
- No regression to the impersonation / session-swap bootstrap logic in `+layout.svelte`.
- Covered by a test (component or E2E) asserting the navbar updates post-login without reload.

## Note (secondary, demo only)
The "semi-broken User" seen on day-2 demo visits is a related-but-separate stale-cookie cold-load symptom (cookie present, backend identity gone after reset). Worth a look while in here, but the primary, always-reproducible bug is the post-login navbar staleness above.

biagiodistefano:
## Summary

Fixes #485 — after logging in, the navbar kept rendering the logged-out chrome (Login / Sign Up) until a manual refresh.

The login (and 2FA) server actions set the auth cookies and `throw redirect(303)`. The enhanced forms followed that redirect as a **client-side** navigation, which does **not** re-run the root layout's server `load` (`src/routes/+layout.server.ts`) — so its `auth.hasAccessToken` flag stayed `false`, the auth-bootstrap `$effect` in `+layout.svelte` never fired, and the navbar stayed logged-out until a hard reload re-ran SSR with the cookie present.

This is the mirror image of logout, which already works precisely because it does a full-page redirect (`/logout` → `redirect(303, …)`).

## The fix

Follow the post-login redirect with a **full-page navigation** (`window.location.href = result.location`), mirroring the logout flow. SSR then re-runs with the freshly-set cookie, the in-memory auth store bootstraps via `/api/auth/refresh`, and the navbar reflects the authenticated session immediately — reusing the already well-tested cold-authenticated-load path.

Non-redirect results (validation errors, the 2FA-required step) still apply **in place** via `applyAction`, so there's no reload on those.

As part of the change, the three byte-identical inline `use:enhance` callbacks (demo / standard / 2FA forms) are consolidated into one shared `handleSubmit: SubmitFunction`.

## Why a full reload (and not `invalidateAll`)

Login is a once-per-session action and logout already full-reloads, so the UX cost is negligible. A full reload reuses the existing, battle-tested SSR bootstrap path rather than relying on subtle client-side invalidation timing around a redirect result. (Alternative `depends('app:auth')` + `invalidateAll()` approaches were considered in the issue; this is the lowest-risk, symmetric one.)

## Testing

- Added `tests/e2e/login-navbar.spec.ts`: logs in via the demo dropdown and asserts the **User menu** is visible and the **Login** link is gone immediately after login, **without** a `page.reload()`. Skips on Firefox (slow cold dev-server load) and when the backend isn't in demo mode — same pattern as `recurring-series-dashboard.spec.ts`.
- `make types` clean for the touched files (the one remaining error is a pre-existing `PUBLIC_VERSION` env issue in `Footer.svelte`, unrelated).
- `prettier --check` and `eslint --max-warnings 0` pass on both changed files.

## Manual verification steps

1. Log in (password, demo account, or 2FA).
2. On landing at `/dashboard`, the navbar shows the user menu/notifications **without** refreshing.
3. Bonus (demo flow): with a stale prior session, Logout → Login — navbar updates immediately, no refresh needed.

### DIFF
--- a/src/lib/utils/safe-redirect.test.ts
+++ b/src/lib/utils/safe-redirect.test.ts
@@ -1,0 +1,38 @@
+import { describe, it, expect } from 'vitest';
+import { safeReturnUrl } from './safe-redirect';
+
+describe('safeReturnUrl', () => {
+	it('accepts genuine relative paths', () => {
+		expect(safeReturnUrl('/dashboard')).toBe('/dashboard');
+		expect(safeReturnUrl('/org/acme/admin/members')).toBe('/org/acme/admin/members');
+		expect(safeReturnUrl('/account/notifications?tab=email')).toBe(
+			'/account/notifications?tab=email'
+		);
+	});
+
+	it('rejects absolute URLs', () => {
+		expect(safeReturnUrl('https://evil.com')).toBe('/dashboard');
+		expect(safeReturnUrl('http://evil.com/path')).toBe('/dashboard');
+	});
+
+	it('rejects protocol-relative and backslash variants', () => {
+		expect(safeReturnUrl('//evil.com')).toBe('/dashboard');
+		expect(safeReturnUrl('/\\evil.com')).toBe('/dashboard');
+	});
+
+	it('rejects non-path schemes and non-relative values', () => {
+		expect(safeReturnUrl('javascript:alert(1)')).toBe('/dashboard');
+		expect(safeReturnUrl('dashboard')).toBe('/dashboard');
+	});
+
+	it('falls back for empty / nullish input', () => {
+		expect(safeReturnUrl(null)).toBe('/dashboard');
+		expect(safeReturnUrl(undefined)).toBe('/dashboard');
+		expect(safeReturnUrl('')).toBe('/dashboard');
+	});
+
+	it('uses a custom fallback when provided', () => {
+		expect(safeReturnUrl(null, '/')).toBe('/');
+		expect(safeReturnUrl('//evil.com', '/home')).toBe('/home');
+	});
+});
--- a/src/lib/utils/safe-redirect.ts
+++ b/src/lib/utils/safe-redirect.ts
@@ -1,0 +1,19 @@
+/**
+ * Resolve a post-auth redirect target, constrained to same-origin relative
+ * paths.
+ *
+ * Guards against open redirects via a crafted `?returnUrl=…`. This matters
+ * because a successful login follows `returnUrl` with a full-page
+ * `window.location.href` navigation on the client (see
+ * `routes/(public)/login/+page.svelte`), so an unvalidated absolute or
+ * protocol-relative value would bounce the freshly-authenticated user off-site.
+ *
+ * Accepts only values that start with a single `/` that is NOT followed by `/`
+ * or `\` — i.e. genuine relative paths. Rejects absolute URLs
+ * (`https://evil.com`), protocol-relative URLs (`//evil.com`), backslash
+ * variants browsers normalise to protocol-relative (`/\evil.com`), and
+ * non-path schemes (`javascript:…`). Anything rejected falls back to `fallback`.
+ */
+export function safeReturnUrl(raw: string | null | undefined, fallback = '/dashboard'): string {
+	return raw && /^\/(?![/\\])/.test(raw) ? raw : fallback;
+}
--- a/src/routes/(public)/login/+page.server.ts
+++ b/src/routes/(public)/login/+page.server.ts
@@ -12,6 +12,7 @@ import { claimPendingTokens, setClaimFlashCookie } from '$lib/server/token-claim
 import { log } from '$lib/server/logger';
 import { buildSeo } from '$lib/seo';
 import { resolveLang } from '$lib/seo/server';
+import { safeReturnUrl } from '$lib/utils/safe-redirect';
 
 export const load: PageServerLoad = ({ url, request }) => {
 	const lang = resolveLang(request);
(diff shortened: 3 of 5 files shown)


biagiodistefano:
### Code review pass (self-review via reviewer subagent)

Ran a senior-reviewer pass over the diff. The core fix was assessed correct across all three login paths; two findings were actioned in 82776f8:

- **Critical — E2E test would fail on mobile.** The regression test asserted on the desktop-only `User menu` button (`hidden md:block` in `Header.svelte`) but only skipped Firefox, so it failed on the `Mobile Chrome` / `Mobile Safari` projects, and the `Login`-link assertion passed trivially there. Now scoped to desktop viewports (`skip` on `isMobile`); the bootstrap path under test is viewport-independent.
- **Security — open redirect, worsened by this change.** The post-login redirect now hard-navigates via `window.location.href`, so an absolute/protocol-relative `?returnUrl=` would bounce the authenticated user off-site (the previous client-side `goto` refused external URLs). Added a unit-tested `safeReturnUrl` util that constrains `returnUrl` to same-origin relative paths, applied to both the login and 2FA actions.

Non-blocking notes accepted as-is: swapping `update()` → `applyAction()` deliberately drops `invalidateAll()`/form-reset (correct here — errors and the 2FA-step `form` data should persist), and `error`-type results render the nearest `+error` boundary as before.

Verification: `safe-redirect` unit test green (6/6); `prettier --check` + `eslint --max-warnings 0` clean on all changed files; `svelte-check` clean for the touched files (the lone `PUBLIC_VERSION`/`Footer.svelte` error is pre-existing and unrelated). The E2E needs a demo-mode backend to run.

PR #491, opened 2026-06-22
fix: hide feature-flag-gated UI when disabled — create-org, Telegram (#488)

### LINKED ISSUE #488: Hide feature-flag-gated UI when the capability is disabled (create-org, Telegram)
## Summary

The backend now gates several capabilities behind `FEATURE_*` flags (letsrevel/revel-backend#562) for self-hosted instances. The frontend has **zero feature-flag awareness** and renders these paths unconditionally, so on an instance that disables a feature, users click into a path that hard-fails with a `403`/`404` and no graceful fallback.

We should hide (or disable) the gated UI when the capability is off.

## Depends on

letsrevel/revel-backend#563 — exposes the flags in the anonymous `GET /version` response as a `features` object. This FE work consumes that. (After it merges, run `pnpm generate:api`.)

### 1. Create organization — `features.organization_creation`
Backend `POST /organizations/` → 403 for non-staff when disabled. Currently shown unconditionally / without a flag check in:
- `src/routes/(auth)/create-org/+page.svelte` — the full create-org form (also `+page.server.ts` calls `organizationCreateOrganization()`)
- `src/lib/components/common/UserMenu.svelte` — "Create Organization" menu link (gated only on `!ownsOrganization()`)
- `src/routes/(auth)/dashboard/+page.svelte` — "Create Organization" CTA (gated only on `!ownsOrganization()`)
- `src/routes/(public)/+page.svelte` — "Start Organizing" CTA on the public home

> Note: the backend lets **staff** bypass the flag. Decide whether to still surface create-org for staff users, or hide for everyone and rely on the server bypass.

### 2. Telegram account linking — `features.telegram`
Backend telegram endpoints → 404 when disabled. Shown unconditionally in:
- `src/routes/(auth)/account/profile/+page.svelte` — renders `<TelegramConnectionManager>` for all authed users
- `src/lib/components/profile/TelegramConnectionManager.svelte` — connect/disconnect/OTP/bot-name UI
- (related) `src/routes/(auth)/org/[slug]/admin/settings/+page.svelte` — org Telegram field; also moot when Telegram is off

### 3. Google SSO — `features.google_sso`
No "Sign in with Google" button exists in the UI today, so nothing to hide right now. **When one is added**, gate it on `features.google_sso` (the flag now defaults to `False`).

## Suggested approach

- Fetch `GET /version` at app boot (layout load / a config store) and expose `features` via a small `$state`/store.
- Conditionally render the paths above. Keep the existing `ownsOrganization()` checks; add the feature-flag check alongside.
- Server-side gating remains the source of truth — this is UX, not a security boundary.

## Refs
- Backend gating + flags: letsrevel/revel-backend#562
- Flags endpoint (prerequisite): letsrevel/revel-backend#563

biagiodistefano:
## Summary

Makes the frontend feature-flag-aware so it hides/disables UI for backend capabilities that are turned off, instead of letting users click into a `403`/`404` dead end. Consumes the `features` object now exposed in the anonymous `GET /version` response (backend letsrevel/revel-backend#563).

Closes #488

## How it works

- **Flags via SSR layout data.** A cached (~5 min TTL), fail-open `getFeatures()` server helper fetches `/version` in the root `+layout.server.ts` and exposes `features` in layout data, so every page/component reads `$page.data.features` during SSR and client navigation **with no flash** (important for the public "Start Organizing" CTA, which is SSR/SEO-visible). The backend stays the security source of truth — this is UX only.
- **Fail-open:** a missing/failed `/version` treats capabilities as enabled, preserving current behavior. The cache is only written on success, so a transient `/version` error never pins a stale state.

## Gated surfaces

- **Create organization** (`features.organization_creation`) — hidden for everyone when off: UserMenu (mobile + desktop), dashboard CTA, public-home "Start Organizing" CTA, and the `/create-org` route is server-guarded (deep-link/bookmark → redirect to `/dashboard`). Backend still bypasses the flag for staff; the FE does not special-case staff (decision below).
- **Telegram** (`features.telegram`) — profile page shows a disabled card with an i18n note (en/de/it) instead of the connect UI; org-admin settings hides the Telegram field. The settings form action was also guarded so a hidden field can't wipe a previously-saved `telegram_url`.
- **Google SSO** (`features.google_sso`) — no SSO button exists yet; documented as a no-op to gate when one is added (flag defaults to `false`).

## Decisions

- **Hide create-org for everyone** when off (no staff special-case in the FE) — simpler; staff can still use the API directly.
- **Telegram → disabled note**, not a blank hole.
- **Deep-link create-org → server-guarded redirect** (no flash, handles bookmarks).

## Tests / checks

- New unit tests: `resolveFeatures` (5) + `getFeatures` cache/fail-open (4) — 9/9 passing.
- `make check`: `format-check`, `types`, `i18n-check` pass; no new `lint`/`file-length` findings (the repo's `lint`/`file-length` baselines are pre-existing red on `main`).
- Full component-test suite not run here (known pre-existing red baseline on `main` from the svelte-query v6 bump, unrelated to this branch).

## Follow-up (non-blocking)

`$page.data.features` is currently untyped (`app.d.ts` leaves `PageData` intentionally untyped); a strict global `PageData` would break other `$page.data.*` accesses, so it was left as-is. All current consumers are safe because the root layout always returns `features`.

## Summary by CodeRabbit

* **New Features**
  * Added feature flag support to control availability of Telegram integration and organization creation per instance.
  * Localized unavailable messaging for German, English, and Italian.

* **Bug Fixes**
  * Fixed organization settings form to prevent unintentional overwrites of Telegram configuration.

* **Tests**
  * Added test coverage for feature flag resolution and caching behavior.


biagiodistefano:
### Code review

PR #571, opened 2026-07-08
ci: close the local-vs-CI QA gap — missing gates, unit tests in CI, green test baseline

biagiodistefano:
Closes the gaps found in the FE-vs-BE QA posture sanity check. Refs #569 (component observations filed during the baseline fix).

## Problem

`make check` and CI had drifted: three gates ran only locally (`file-length`, `no-ssr-token`, `audit-images`), **no tests ran in CI at all**, and the main ruleset had zero required status checks — so the component suite sat red on `main` for weeks (168 failing tests post svelte-query v6) with nothing to flag it.

### CI (`ci.yml`)
- Lint job now also runs the three previously local-only gates: file-length, SSR access-token leak guard, image-alt audit
- New **Unit Tests** job: `pnpm install` → `paraglide:compile` → `vitest run`

### `deps-noop.yml` (new)
No-op twin of `deps.yml` — the backend's documented pattern for path-filtered required checks. Reports instantly-passing jobs under the same context names ("pnpm audit (prod, high+)", "License compliance") on PRs that don't touch the dependency graph, so those checks can be required without hanging PRs. **Invariant:** its `paths-ignore` must stay the exact complement of `deps.yml`'s `paths`.

### Test baseline: 168 failed → 0 (test-only changes, no product code)
- **9 files (~125 tests)**: migrated off the pre-v6 `render(QueryClientProvider, {children: X})` pattern to the existing `QueryClientTestWrapper`
- **French locale (#484)**: hreflang/resolveLang tests updated for the 4-language reality (fallback path still covered via a genuinely unsupported language)
- **Stale assertions**: moved features (ticketing/city steps), renamed i18n copy, a11y-pass DOM changes (#564/#567), `EventStatus` enum drift (`approved`→`open` etc.), series route change, past-date fixture time bombs (now relative dates)
- **Test-infra fixes**: vitest 4 `spyOn`-on-prototype-mock trap (scroll tests), ESM-incompatible `require()` calls, jsdom `body.style` leakage between bits-ui dropdown tests
- Two low-severity component observations found during the fix were deliberately **not** patched here — filed as #569 (TagInput label association, EventStatusBadge Full/Past priority)

## Ruleset (applied via API, not in this diff)

The `main` ruleset gains BE-parity enforcement: 7 required status checks (Lint & Type Check, Unit Tests, Translation Validation, Build, Secret scan (gitleaks), pnpm audit, License compliance), strict up-to-date policy, and required signatures.

## Verification

[...]

PR #572, opened 2026-07-08
fix(components): TagInput label association + EventStatusBadge Full/Past priority

### LINKED ISSUE #569: Two component soft spots surfaced by the test-baseline fix: TagInput label association, EventStatusBadge Full/Past priority
While fixing the red vitest baseline for the CI/QA-parity PR, two low-severity product observations surfaced. Per the fix-tests-not-components rule, neither was touched — documenting here instead.

## 1. `TagInput.svelte` — broken label/input association when no `id` prop is passed

`src/lib/components/forms/TagInput.svelte` renders `<label for={inputId}>` (auto-generated fallback id) but puts the raw `{id}` prop on the `<input>`. When the consumer omits the `id` prop, the input has **no id at all**, so the label is associated with nothing — a WCAG 1.3.1/4.1.2 issue (screen readers announce an unlabeled input; clicking the label doesn't focus the field).

**Fix:** the input should use `id={inputId}` (the fallback-aware value), not `{id}`.

**Mitigating fact:** grep found **no production usages** of this component — it's currently an orphan (the admin forms use their own tag inputs). So this is latent, not user-facing. Consider either fixing the one-liner or deleting the orphan component.

The tests currently work around it by always passing an explicit `id: 'tags'` (with a comment saying so — see `TagInput.test.ts`). If the one-liner is fixed, that workaround can be dropped.

## 2. `EventStatusBadge.svelte` — doc comment contradicts code on Full vs Past priority

The component's own doc comment lists badge priority as **Full > Today > Ongoing > Past**, but the code checks capacity before temporal status, i.e. **Full > Past > Ongoing > Today**. Net effect: an event that is both full and already ended shows **"Full"** rather than "Past", which is arguably wrong (the event being over is more relevant than it having been full).

**Fix options:** (a) check temporal end-state before capacity so ended events show "Past", and align the comment; or (b) decide current behavior is intended and fix the comment only. The test suite currently pins the shipped behavior ("Full over Past" — see `EventStatusBadge.test.ts`), so whichever way this lands, update that test alongside.

---

Origin: vitest-baseline repair (svelte-query v6 fallout + stale assertions) feeding the CI hardening PR that adds a required Unit Tests job. Refs #484 (fr locale), the a11y passes in #564/#567.

biagiodistefano:
Closes #569

Two low-severity component soft spots surfaced during the vitest-baseline repair for #571, deliberately left untouched there (fix-tests-not-components rule) and filed as #569. This PR fixes both.

## 1. `TagInput.svelte` — label/input association broken without an `id` prop

The component rendered `<label for={inputId}>` (fallback-aware auto-generated id) but put the raw `{id}` prop on the `<input>`. With no `id` prop, the input had no id at all, so the label pointed at nothing — WCAG 1.3.1/4.1.2 (unlabeled input for screen readers; clicking the label didn't focus the field).

**Fix:** the input now uses `id={inputId}`. The stale test-side note explaining the always-pass-an-id workaround is removed, and a regression test asserts `getByLabelText` resolves to the input with no `id` prop passed.

Note: the component is currently an orphan (no production usages), so this is latent hardening. Kept rather than deleted since it's a complete, tested, accessible form control.

## 2. `EventStatusBadge.svelte` — Past now wins over Full

The doc comment promised **Full > Today > Ongoing > Past**, but the code checked capacity before temporal status, so an event that was both full and already ended showed **"Full"**. Per the issue's reasoning, the event being over is more relevant than it having been at capacity, so the Past check now runs before the capacity check (option (a) from the issue).

- Doc comment and `EventStatusBadge.showcase.md` aligned with the actual priority: Cancelled/admin > Past > Full > Ongoing > Today > Upcoming.
- The priority-pinning test flipped to assert "Past over Full".
- The two plain "Full" tests now pin `vi.setSystemTime` to a moment while the mock event is still upcoming — they only passed before because Full short-circuited ahead of Past for an event date that is now in the real past.

## Verification

- `pnpm vitest run` (full suite): 83 files, 814 passed, 1 todo
- `make check`: green (format, lint, types, i18n, file-length, no-ssr-token, audit-images)

## Ride-along: showcase doc staleness

The code review surfaced pre-existing stale facts in `EventStatusBadge.showcase.md`: it described Cancelled as `status === 'rejected'` with a red XCircle badge, but the component checks `'cancelled'` and renders the orange `cancelled` variant with the `Ban` icon (XCircle/red belong to Closed). Corrected in a follow-up commit.

### DIFF
--- a/src/lib/components/events/EventStatusBadge.showcase.md
+++ b/src/lib/components/events/EventStatusBadge.showcase.md
@@ -16,8 +16,8 @@ A visual indicator that displays the current temporal status of an event with se
 The badge displays one of the following statuses based on priority:
 
 1. **Cancelled** (Red/Destructive) - Event status is 'rejected'
-2. **Full** (Red/Destructive) - Event has reached maximum capacity
-3. **Past** (Gray/Secondary) - Event has ended
+2. **Past** (Gray/Secondary) - Event has ended (wins over Full: an ended event is over, regardless of capacity)
+3. **Full** (Red/Destructive) - Event has reached maximum capacity
 4. **Ongoing** (Green/Success) - Event is currently happening
 5. **Happening Today** (Green/Success) - Event starts today but hasn't started yet
 6. **Upcoming** (Blue/Default) - Event is scheduled for the future
@@ -85,16 +85,16 @@ The component uses the following logic to determine status (in priority order):
 event.status === 'rejected';
 ```
 
-### 2. Full
-
-```typescript
-event.max_attendees > 0 && event.attendee_count >= event.max_attendees;
-```
-
-### 3. Past
+### 2. Past
 
 ```typescript
 new Date(event.end) < new Date();
+```
+
+### 3. Full
+
+```typescript
+event.max_attendees > 0 && event.attendee_count >= event.max_attendees;
 ```
 
 ### 4. Ongoing
--- a/src/lib/components/events/EventStatusBadge.showcase.md
+++ b/src/lib/components/events/EventStatusBadge.showcase.md
@@ -15,7 +15,7 @@ A visual indicator that displays the current temporal status of an event with se
 
 The badge displays one of the following statuses based on priority:
 
-1. **Cancelled** (Red/Destructive) - Event status is 'rejected'
+1. **Cancelled** (Orange) - Event status is 'cancelled'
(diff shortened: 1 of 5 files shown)


biagiodistefano:
### Code review

PR #615, opened 2026-07-12
test(e2e): Phase 3b — j01/j02/j12/j15/j21 @p2 guest & public edges

biagiodistefano:
Part of #591 (E2E suite v2 — Phase 3, @p2). Second grouped PR; independent of #614 (group a).

## New @p2 journey specs (10 tests, green on chromium + mobile-chrome, `--retries=0 --repeat-each=3`, zero flakes)

- **j01/legal-marketing** — `/legal/privacy`, `/legal/terms`, and a marketing landing page each render (200 + level-1 heading).
- **j01/contact-org** — arrange an own public org with `contact_method=form`; a signed-in visitor sends a message and the organizer's mailbox receives it (subject `[{org}] {subject}`).
- **j02/referral-code-registration** — `?ref=B2CREF01` auto-applies ("Referral code applied") and registration lands on the check-email page; a bogus code is rejected inline.
- **j12/token-expiry** — the two 410 quadrants `j01/token-preview` didn't cover: an **expired org** token and a **used-up event** token, each showing the reason and no "Sign In to Claim".
- **j15/unsubscribe** — an invitation email's tokened `/unsubscribe` link (no auth) → save → "Preferences Updated".
- **j21/referral** — the seeded B2B referrer (`referrer.b2b@`, password `password`) sees code `B2BREF01` on `/account/referral` and a **Paid** payout with a self-billing statement + Download PDF on `/account/referral/payouts`.

## Gate

- Group specs: **60 passed** (both projects × repeat-each=3), zero flakes.
- Fresh-DB full suite: **235 passed / 2 mobile-viewport skips / 0 flaky** (branch tally 217 → 237).
- `make check` ✅ (no src/support-kit changes, so the unit suite is unaffected).

## Notes / drift

- **Referrer accounts use password `password`**, not the persona bootstrap `password123`; both `/account/referral*` routes redirect non-referrers to `/dashboard`, so this runs as an actual referrer.
- "Verify referral via referrer payout page" (the design's wording) isn't achievable for a *new* registration — the `Referral` row is created immediately but no `ReferralPayout` (payouts only exist for periods with real payments) and there's no referred-users surface. Spec asserts the FE-observable applied/rejected states + a successful submit instead.
- Mobile gotcha encoded: the mobile nav drawer is itself `role="dialog"` and carries section headings, so headings/dialogs are scoped (`level: 1`, dialog `name`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

[...]

PR #614, opened 2026-07-12
test(e2e): Phase 3a — j03/j17 @p2 account & lifecycle suites

biagiodistefano:
Part of #591 (E2E suite v2 — Phase 3, @p2). First of the grouped PRs; groups b–f follow.

## New @p2 journey specs (16 test instances, green on chromium + mobile-chrome, `--retries=0 --repeat-each=3`, zero flakes)

**j03 — account basics**
- `dietary` — add preference (private) + restriction backed by a newly-created food item; visibility toggle; remove via native `confirm()`.
- `notification-prefs` — silence-all disables dependent controls; drop the email channel; daily digest + send-time; persistence across reload (bits-ui checkbox/radio controls).
- `billing-profile` — create the buyer billing profile and confirm it persists (form flips to "Update", VAT-ID section appears).

**j17 — account lifecycle** (all register throwaway users — these mutate/destroy the login identity)
- `2fa` — enable TOTP from the exposed base32 secret (adds `otplib` devDep), prove login now requires the second 6-digit step, disable.
- `email-change` — request with password → old address notified + new address confirms → the confirming session survives (fresh token pair) → only the new email logs in.
- `gdpr-export` — request → emailed signed backend link → direct fetch returns the export (200, non-empty).
- `account-deletion` — Danger-Zone modal → emailed confirmation → account gone, credentials dead.

## Support kit & in-PR fixes

- **`fetchWithRetry`** extracted in `support/api.ts`; the unauthenticated arrange calls (`register`/`verify`/guest RSVP + confirm) now use it, matching the bounded 5xx retry `ApiClient` already had (#610). These previously hard-failed on the dev backend's Postgres *"sorry, too many clients already"* transient under parallel load; now they ride over it.
- **Pre-existing @p0 flake fixed** in `login-logout` (surfaced by the full-suite run): `scrollIntoViewIfNeeded` on the mobile drawer's Log Out / Login button raced the auth-bootstrap re-render (`element is not stable`). Both scroll sites now retry the scroll+viewport assertion together, re-resolving the locator each pass — the same pattern `openMobileNav` already uses. Verified 12× stable on mobile-chrome.

## Gate

- Group specs: **48 passed** (both projects × repeat-each=3), zero flakes.
- Fresh-DB full suite: **230 passed / 2 mobile-viewport skips / 0 flaky** (tally 217 → 233).
- `make check` ✅ · `make test` (844) ✅

## Notes / drift

[...]

Issue #593, opened 2026-07-10
E2E v2 — CI wiring


biagiodistefano:
Part of the E2E test suite v2 (umbrella: #28). Design: `docs/superpowers/specs/2026-07-10-e2e-test-suite-design.md` (§8). Requires Phase 1 (meaningful once @p0 is green).

- [ ] `e2e.yml`: Postgres/PostGIS + Redis services; checkout backend, `uv sync`, `make bootstrap`, Django with `CELERY_TASK_ALWAYS_EAGER=1`; Mailpit container
- [ ] PR tier: `--grep-invert @p3`, excluding `@stripe`-tagged specs; target @p0+@p1 < ~10 min
- [ ] Nightly tier: full set with `STRIPE_SECRET_KEY` + `stripe listen` (evaluate stripe-mock as an alternative)
- [ ] HTML report + traces uploaded as artifacts on failure
- [ ] Decide backend ref pinning strategy (main vs pinned SHA)

biagiodistefano:
# E2E CI Wiring Design (#593)

Closes the last phase of the E2E suite v2 (umbrella #28, design
`2026-07-10-e2e-test-suite-design.md` §8). Supersedes §8's PR-tier/nightly
sketch — the strategy below was decided 2026-07-14.

## 1. Strategy

**E2E never blocks day-to-day PRs.** The suite gates **shipping**: it runs on
release PRs (`release/v*`), full set, against the backend image that
production actually runs. Rationale:

- FE and BE are developed in parallel; FE `main` is routinely ahead of the
  released BE (e.g. the two-step checkout window, FE #646 merged while BE
  #702 was unreleased). E2E-on-every-PR against the released BE would be
  red for legitimate reasons often enough to train everyone to ignore it.
- **Release-first discipline**: a BE release must be cut before an FE release
  that depends on it. A red gate on the release PR is an honest "prod would
  break" signal, caught before anything irreversible (no tag, no image, no
  merge).
- No nightly tier for now (decided 2026-07-14; can be added later by giving
  the reusable workflow a `schedule` caller).

The cheap `regression/` specs (CSP/FOUC, frontend-only, 5 browsers) DO run on
every PR as a new job in `ci.yml` — they need no backend and finish in ~2 min.

## 2. Backend provisioning

- Image: `ghcr.io/letsrevel/revel:${{ vars.E2E_BACKEND_IMAGE_TAG || 'latest' }}`.
[...]

PR #620, opened 2026-07-12
fix(e2e): transient network flakiness, clock drift, and test quality improvements

### LINKED ISSUE #619: E2E audit: transient network flakiness, clock drift, and test quality improvements (PR #614 & #615)
A comprehensive audit of the E2E test changes in PR #614 and PR #615 identified 5 robustness issues and 4 code quality improvements. These span missing transient error handling, raw un-retried network calls on overloaded endpoints, clock-drift sensitivity, code duplication, and locator/assertion Page Object gaps.

Each finding has an isolated, cherry-pickable fix in PR #620.

---

### 1. `fetchWithRetry` Network Exceptions Blind Spot

**Introduced in**: PR #614
**File**: `tests/e2e/support/api.ts`

#### Problem

The helper `fetchWithRetry` retries HTTP status codes `>= 500` but does not catch network-level exceptions thrown by the native `fetch()` implementation (e.g., `TypeError: fetch failed` due to socket hang-ups, connection aborts, DNS timeouts, or connection reset errors). Furthermore, its backoff delay was not abort-aware, meaning if the request's `AbortSignal` was triggered during the `setTimeout` backoff delay, the worker would keep waiting until the backoff timeout completed.

#### Fix

Wrap the native `fetch` in a `try/catch` block, rethrowing the exception immediately if the `AbortSignal` is aborted or if all retry attempts are exhausted. Additionally, make the backoff delay abort-aware by listening to `init.signal` abort events and rejecting immediately.

Fixed in [`ab561d5`](https://github.com/letsrevel/revel-frontend/pull/620/commits/ab561d5246bdf3521eac306158301d916f794120) (PR #620).

---

### 2. Code Duplication in `ApiClient.request`

**Introduced in**: PR #614
**File**: `tests/e2e/support/api.ts`

#### Problem

`ApiClient.request` duplicates the attempt-tracking, delay-backoff, and status-checking retry loop of `fetchWithRetry`, violating DRY principles and leaving the client method vulnerable to uncaught network-level exceptions.

#### Fix

Refactor `ApiClient.request` to delegate HTTP execution directly to the corrected `fetchWithRetry` helper.

Fixed in [`f9dcab2`](https://github.com/letsrevel/revel-frontend/pull/620/commits/f9dcab22c02c741630658ba550bffeb7a967ae65) (PR #620).

---

### 3. Missing `fetchWithRetry` in Backend Media Downloads

**Introduced in**: PR #614 & PR #615
**Files**:
- `tests/e2e/journeys/j17-account-lifecycle/gdpr-export.spec.ts`
- `tests/e2e/journeys/j21-referral/referral.spec.ts`

#### Problem

Both tests call the native global `fetch` directly to download files served by the backend (a GDPR ZIP export and a PDF payout statement). Under parallel test suite runs, these endpoints can return transient 500/502 errors due to container resource contention. Additionally, the referral PDF download flow failed to resolve relative URLs through `getBackendUrl` before downloading, which differed from the main payouts page behavior.

#### Fix

Import and use the `fetchWithRetry` helper for both requests, resolving the referral statement download URL using the new `getBackendUrl` helper beforehand to support both relative and absolute URLs.

Fixed in [`f70e446`](https://github.com/letsrevel/revel-frontend/pull/620/commits/f70e446ecb38c7387f66298db5cb472308e4887f) (PR #620).

---

### 4. Clock/Time Drift Risk in Token Expiry Backdating

**Introduced in**: PR #615
**File**: `tests/e2e/journeys/j12-invitations-tokens/token-expiry.spec.ts`

#### Problem

The test backdates the organization token's expiration by only 60 seconds (`expires_at: new Date(Date.now() - 60_000).toISOString()`). In virtualized or containerized (WSL2, Docker Compose, or CI VM) test environments, container clocks often drift and lag behind the host/test runner clock. If the backend clock lags by more than 60 seconds, it evaluates the token as still active, causing the test to fail.

#### Fix

Change the backdating offset to 1 hour (matching the safer margin used in `token-preview.spec.ts`) to eliminate clock drift risk.

Fixed in [`ec31f09`](https://github.com/letsrevel/revel-frontend/pull/620/commits/ec31f0933832e6ccaed24d4bfa409cada755d870) (PR #620).

---

### 5. Backend Probe Missing Retry in `skip.ts`

**Introduced in**: PR #614
**File**: `tests/e2e/support/skip.ts`

#### Problem

The backend availability check uses the native `fetch` to test the `/api/version` endpoint with a single try. If this request fails or times out under a transient parallel container startup load, the test worker caches a `null` version info, resulting in the worker silently skipping *all* specs in the suite.

#### Fix

Update `skip.ts` to use `fetchWithRetry` to handle transient connection failures during initial worker setup.

Fixed in [`a06ca2a`](https://github.com/letsrevel/revel-frontend/pull/620/commits/a06ca2a1040eb9447572fc53377ee1e918086509) (PR #620).

---

### 6. Missing Reply-To Assertion & Inline Selectors in Contact Form Test

**Introduced in**: PR #615
**File**: `tests/e2e/journeys/j01-guest/contact-org.spec.ts`

#### Problem

The contact form test asserts the email's Subject but does not check the `Reply-To` header. Furthermore, the contact organizer flow interacted directly with low-level page locators and dialog elements rather than using a reusable Page Object.

#### Fix

Extract the contact form and dialog selectors into a reusable `ContactDialogPage` Page Object class, and add a `ReplyTo` assertion verifying the email's Reply-To address points to the sender.

Fixed in [`25701da`](https://github.com/letsrevel/revel-frontend/pull/620/commits/25701da4db68069cc606ca54ddf1e5ec7cce3063) (PR #620).

---

### 7. Inline Registration Form Locators

**Introduced in**: PR #615
**File**: `tests/e2e/journeys/j02-registration/referral-code-registration.spec.ts`

#### Problem

The registration-code spec interacts directly with low-level page locators (inputs, checkboxes, and buttons) inside a complex `toPass` retry loop. Additionally, the terms checkbox interaction used an unnecessary `isChecked()` check before calling `.check()`.

#### Fix

Extract the registration form locators and flows into a reusable `RegisterPage` Page Object class, decoupling selector structure from test logic, and simplify terms checkbox checking by calling `.check()` directly.

Fixed in [`bf58cf1`](https://github.com/letsrevel/revel-frontend/pull/620/commits/bf58cf10ca2e28a289ef92b59b4ed2085d5f2e24) (PR #620).

---

### 8. Brittle Positional Locator in Unsubscribe Flow

**Introduced in**: PR #615
**File**: `tests/e2e/journeys/j15-notifications/unsubscribe.spec.ts`

#### Problem

The unsubscribe page has a single form, but the test uses `.last()` to locate the "Save Changes" button — a pattern carried over from settings pages that have multiple forms. The positional selector is unnecessary and fragile if the page layout changes.

#### Fix

Scope the button locator to the relevant card container (`.bg-card`), eliminating positional reliance.

Fixed in [`1a23908`](https://github.com/letsrevel/revel-frontend/pull/620/commits/1a2390878355465a6db03f6444b01d26ebbb4e41) (PR #620).

---

### 9. Inline Referral Page Locators

**Introduced in**: PR #615
**File**: `tests/e2e/journeys/j21-referral/referral.spec.ts`

#### Problem

The referral program spec directly accesses DOM elements for navigation, referral-code validation, selecting payout rows, opening the self-billing invoice statement dialog, and trigger pdf download. This makes the test code verbose and tightly coupled to the UI structure.

#### Fix

Extract referral program page selectors and actions into a reusable `ReferralPage` Page Object class.

Fixed in [`39c7a32`](https://github.com/letsrevel/revel-frontend/pull/620/commits/39c7a320ffb58aa78ab26ca998255fab2e245db9) (PR #620).

florimond-collette:
Follow-up to the Phase 3 E2E PRs (#614 and #615), introducing robust transient failure handling, clock drift safety, DRY refactoring, and test quality improvements.

Closes #619.

Each fix is in a separate commit, making them easy to cherry-pick individually.

### Robustness Fixes
1. **Catch network exceptions and abort early in `fetchWithRetry`**: wraps native `fetch` in `try/catch` to handle transient socket failures, and makes the backoff delay abort-aware by listening to `init.signal` abort events.
2. **Refactor `ApiClient.request`**: reuse the central `fetchWithRetry` helper inside the API client, eliminating duplicated retry logic.
3. **Use `fetchWithRetry` in GDPR and Referral downloads**: makes unauthenticated static asset downloads resilient to transient 500s under parallel worker load, resolving relative URLs via the new `getBackendUrl` helper first.
4. **Token expiry drift protection**: backdates token expiry by 1 hour (instead of 60 seconds) to avoid tests failing under CI clock drift or lag.
5. **Backend version probe retry**: wraps the worker startup availability check with `fetchWithRetry` to prevent workers from silently skipping all specs on a transient startup failure.

### Code Quality Improvements
6. **Reply-To header assertion**: extends the contact-org test to assert that the received email's Reply-To header points to the sender.
7. **Scope Save Changes button**: scopes the button locator to `.bg-card` in the unsubscribe journey to avoid brittle positional selection.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when accessing emailed downloads and invitation links, including temporary backend failures.
  * Verified that contact-organizer emails correctly support replies to the sender.
  * Strengthened handling of expired invitation tokens and notification unsubscribe confirmations.

* **Tests**
  * Expanded end-to-end coverage for one-click notification unsubscribing, GDPR exports, referral statements, and email workflows.
  * Improved validation of downloaded files and email preferences.


biagiodistefano (changes requested):
### Requesting changes: keep the fixes, drop the Page Object refactor

Thanks for this — the robustness fixes are exactly right and I'd merge them as-is:

- `fetchWithRetry` now catching **network exceptions** (not just 5xx) with abort-aware backoff, plus wrapping the `/api/version` probe with it — this directly attacks the transient connection-reset / silent-self-skip flakiness we keep hitting.
- `ApiClient.request` reusing `fetchWithRetry` (the DRY cleanup CodeRabbit flagged on #614).
- GDPR/payout downloads via `fetchWithRetry` + the new `getBackendUrl` helper.
- Widening the token backdate to 1h to tolerate clock drift.
- The Reply-To assertion in `contact-org` (+ the `ReplyTo` field on the Mailpit interface).

I ran all six affected specs locally on both projects and they pass — worth noting since CI doesn't run the journey suite yet (#593), so these weren't actually executed by the checks.

### The one change I'd like before merge: please drop the Page Object extraction

That's `tests/e2e/support/pages/{register,referral,contact-dialog}.ts` and the corresponding spec rewrites — restore the inline locators.

**Why:**

1. **It reverses a documented decision.** The suite design spec (§5) explicitly chose *"No Page Objects — plain helper functions per domain; revisit only if duplication actually hurts."* That was a deliberate call, not an accident.
2. **No dedup benefit yet.** Each new page object is used by exactly **one** spec, so it isn't removing cross-spec duplication — the stated trigger for revisiting the decision hasn't been met.
3. **It creates a split-brain on the register form.** `register-verify.spec.ts` still drives registration through the `auth-forms.ts` helpers (`fillRegistrationForm` / `revealRegistrationForm`), while `referral-code-registration.spec.ts` now goes through `RegisterPage`. Two idioms for the same form is precisely the inconsistency the single-style convention exists to prevent.

If we do want Page Objects suite-wide, that's a fair thing to propose — but as its own deliberate change: update spec §5 and migrate **all** specs (including the `auth-forms` helpers) in one pass, rather than half-converting a few specs.

**One entanglement to watch:** the Reply-To assertion is bundled with the `ContactDialogPage` extraction in the same commit, so that one needs a light split — keep the `ReplyTo` assertion and the `MailpitMessageSummary.ReplyTo` interface change, just restore the inline dialog interaction.

Everything else is good to go. 👍

florimond-collette:
Thanks for the detailed review and for running the specs locally, @biagiodistefano! 

To clarify the context behind the Page Object additions and address how we can proceed:

1. **Why they were added:** The Page Object extractions were introduced to address the automated review comments from **@coderabbitai** on previous runs. 
2. **Access to Design Specs:** Adhering to the project's design specs (like §5 of the suite design spec) is currently quite difficult for external contributors because these guidelines are not committed to the repository. Looking at lines 99-102 of `.gitignore`, directories like `.dev-docs/` and `docs/superpowers/` are local-only. We can deduce that the specs should live under `docs/superpowers/specs/`, but because this directory is ignored, it is invisible to external contributors. Crucially, the root `CLAUDE.md` file (line 7) itself references `.dev-docs/guides/common-fixes.md` as an "Important Reference Document" that one should "READ THIS FIRST," reinforcing that this critical documentation is expected to be read but is completely inaccessible to external developers.
3. **Easy Cherry-Picking:** Because the PR was structured with logical, separate commits, it is very easy to cherry-pick only the robustness and functional fixes you want to merge. For example, the `ReferralPage` extraction is isolated in commit `39c7a32`, and the `RegisterPage` extraction is in `bf58cf1`. 

To keep things simple, I will clean up this PR by reverting the Page Object commits and restoring the inline locators directly on this branch, leaving only the robustness and functional fixes ready for merge.

biagiodistefano (approved):

PR #622, opened 2026-07-12
feat(attendees, members): create RSVP on behalf + membership-tier reorder (#616, #618)

biagiodistefano:
Completes the FE work that was backend-blocked in #616 / #618, now that the backend has landed the enabling changes (client regenerated in the first commit).

## #616 — Create RSVP on behalf of a member
Adds an **"Add attendee"** action to the event-admin Manage Attendees page, wired to the existing `eventadminrsvpsCreateRsvp` endpoint via a new `CreateRsvpDialog`.

- Reuses the accessible `MemberCombobox` (searches org members) + a yes/maybe/no status radiogroup mirroring `EditRsvpDialog`.
- The endpoint upserts on `(event, user)`, so submitting for a member who already RSVP'd updates their status — success copy is neutral ("RSVP saved").
- **Scope note:** the picker is **org-members-only**, because that members-list endpoint (opened up to staff in backend #686) is the *only* user-lookup the API exposes — there is no arbitrary-user/email search. Broader "any user by email" lookup would need a further backend change; flagged on #616.
- The multi-select status filter half of #616 already shipped in #621.

## #618 — Reorder membership tiers
Adds **Move up / Move down** controls to each tier card in `TiersTab`, mirroring the existing ticket-tier reorder UX, now that the backend exposes `display_order` + `PATCH .../membership-tiers/reorder` (backend #687).

- Shown only with ≥2 tiers; up disabled on first, down on last; all disabled while a reorder is in flight.
- Optimistic swap on the shared `['organization', slug, 'membership-tiers']` query cache; `invalidateQueries` on success/error resyncs from the server (source of truth).

## Verification
`make check` green — Prettier, ESLint `--max-warnings 0`, `svelte-check` **0 errors**, i18n-check 100% (en/de/fr/it), file-length, SSR-token + image-alt audits. New i18n keys added to all 4 locales. Runtime smoke (create-on-behalf + reorder against a seeded org) pending reviewer.

Closes #616
Closes #618

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #617, opened 2026-07-12
Event admin: no statistics tab (surfaced by E2E j10/stats)


biagiodistefano:
Surfaced while building the @p2 E2E suite (#591, group e — `j10/stats-schedule-resources`).

The E2E design assumed an event-admin **statistics** surface, but **none exists** — no dedicated stats tab/page in the frontend and **no backend statistics endpoint** (the only `stat`-related SDK calls are questionnaire/discount-code stats). The closest thing that renders is the client-side `AttendeeStats` summary (yes/maybe/no + total counts) on the Manage Attendees page.

Also note (related spec drift, not a bug): the schedule editor and event-resources picker are **sections inside the Details tab** of the event editor, not separate tabs — the E2E spec was written accordingly.

**Ask (needs a design decision, then BE + FE):** either
1. build a real event-statistics endpoint + admin tab (RSVP/ticket/check-in/revenue rollups), or
2. formally scope the "statistics" journey down to the existing `AttendeeStats` counts and close this out.

The E2E `j10/stats-schedule-resources` spec is deferred until this is decided.

biagiodistefano:
**Decision (2026-07-12): scope down to the existing `AttendeeStats` surface — issue stays open pending the notes below.**

Chosen option 2 from the ask: do **not** build a new backend statistics endpoint/tab. The event-admin Manage Attendees page already renders `AttendeeStats` (`src/lib/components/attendees/AttendeeStats.svelte`): a **Total** card (accurate server-side `totalCount`) plus **Yes / Maybe / No** breakdown cards. It already self-documents its one limitation — when the RSVP list spans multiple pages, the yes/maybe/no breakdown is computed from the loaded page only, and the component shows an explicit `role=alert` warning (`attendeesAdmin.warningIncompleteData` / `warningTotalRsvps`).

### Next steps
- Point the `j10/stats-schedule-resources` E2E spec at the existing `AttendeeStats` counts. Note the schedule editor + event-resources picker are **sections inside the event editor's Details tab**, not separate tabs — write the spec accordingly.
- If accurate full-event yes/maybe/no breakdowns (independent of pagination) or ticket/check-in/revenue rollups become a real need later, that is a **new backend statistics endpoint** — open a fresh, separately-scoped issue rather than expanding this one.

Leaving open so the E2E spec re-point is tracked; close once that lands (or convert to the new stats-endpoint issue if we go that way).

Issue #591, opened 2026-07-10
E2E v2 — Phase 3: @p2 journey suites


biagiodistefano:
Part of the E2E test suite v2 (umbrella: #28). Design: `docs/superpowers/specs/2026-07-10-e2e-test-suite-design.md` (§7). Requires Phase 2.

- [ ] Tickets: `offline` (staff confirm), `at-the-door`+`pwyc`, `seat-selection`, `self-cancel` (preview + Stripe refund)
- [ ] `j05/waitlist` (join/leave), `j10/manage-rsvps`, `j10/invitation-requests`, `j10/waitlist-admin`, `j10/edit-duplicate`, `j10/stats-schedule-resources`
- [ ] `j01/legal-marketing`, `j01/contact-org`, `j02/referral-code-registration`, `j03/dietary`, `j03/notification-prefs`, `j03/billing-profile`
- [ ] `j04/tier-benefits`, `j04/status-changes`, `j08/membership-tiers`, `j08/blacklist`, `j08/venues`, `j08/resources`, `j08/vat-billing`, `j09/permission-gating`
- [ ] `j11/conditional-branching`, `j11/retake`, `j12/token-expiry`, `j13/hard-blacklist`, `j15/unsubscribe`, `j16/vat-preview`
- [ ] `j17/2fa` (otplib devDep), `j17/email-change`, `j17/gdpr-export`, `j17/account-deletion`
- [ ] `j18/recurring-admin`, `j18/follow-series`, `j20/discount-lifecycle`, `j21/referral`, `j22/buyer-invoices`, `j23/plans-admin`, `j23/subscription-lifecycle`, `j25/financials`

PR #626, opened 2026-07-12
test(e2e): Phase 3e — j05/j10 @p2 event-admin suites + gate-surfaced fixes (#591)

biagiodistefano:
Group **(e) event admin** of the @p2 phase — six new journey suites (8 tests × 2 projects), plus three product fixes and one E2E-infra fix the suites surfaced. Each fix is its own commit for easy cherry-picking.

Part of #591.

## New suites

| Spec | Covers |
|---|---|
| `j05-rsvp/waitlist` | Join + leave the waitlist on a full RSVP event (inline success → self-reload → on-waitlist gate; native confirm on leave) |
| `j10-event-admin/manage-rsvps` | **Create RSVP on behalf** (#622 dialog: members-only combobox, upsert on (event,user)) + **multi-select status filter** (#621, aria-pressed toggles → `?status=a,b`) — doubles as the runtime smoke both PRs deferred |
| `j10-event-admin/invitation-requests` | User requests via dialog on a private event; admin approves/rejects (inline alert banner, **no tier param** — confirmed against the controller); grant proven by requester's RSVP succeeding while the rejected user stays 400 |
| `j10-event-admin/waitlist-admin` | Issue → revoke → reactivate offers + remove entry. Reactivate lives on the **Offers tab behind the Revoked chip** (only PENDING offers surface as `current_offer` on Entries) |
| `j10-event-admin/edit-duplicate` | Rename persists; duplicate → draft copy on the **copy's** edit page |
| `j10-event-admin/stats-schedule-resources` | AttendeeStats tiles as the stats surface (no stats tab exists — #617); schedule session + attached org resource reach the public page for a guest |

## Fixes surfaced by the suites

1. **Stale eligibility gate after invitation/whitelist request on RSVP events** — the `onInvitationRequestSuccess`/`onWhitelistRequestSuccess` → `refreshUserStatus` chain was wired for the ticketed path but stopped at `EventRSVP` on the non-ticketed path, so the gate kept offering "Request Invitation" after a successful request. Threaded the callbacks through.
2. **Duplicate flow edited the wrong event** — `EventEditor` snapshots `existingEvent` (incl. the id it saves to) into `$state` once; the duplicate modal's client-side `goto` reused the mounted editor, so the copy's edit page still showed — and would have saved to — the *original* event. `{#key event.id}` remount.
3. **a11y: mobile attendee cards** — bare "Edit"/"Delete"/"More actions" on every row; now per-user labels mirroring the desktop table (existing i18n keys).
4. **E2E infra: Stripe adaptive pricing** — when Stripe geolocates the runner outside the currency zone, hosted checkout shows a "Choose a currency" variant that broke the card-form probe (both `stripe-online` runs). Role-based probe + pick the original € price.

## Backend issues filed (not touched)

- **revel-backend#691** — `rsvp()` asserts capacity unconditionally, so a YES attendee on a *full* event can't downgrade to no/maybe (exactly when freeing a seat matters; the YES→non-YES waitlist-processing branch is unreachable). Worked around in `waitlist-admin` via admin RSVP delete.

## Gates

- New suites: 16/16 both projects; **48/48** on `--repeat-each=3 --workers=4 --retries=0` against the gunicorn+PgBouncer backend (BE #688) — full parallelism, zero connection pressure (old runserver limits gone; full suite now ~6 min at workers=4).
- **Final full suite: 267 passed / 0 failed / 0 flaky / 2 skipped in 2.8 min** (`--workers=4`, fresh backend). Earlier runs on the way there: 260 passed / 6 failed — all 6 pre-existing environmental (org-list seed pollution from accumulated E2E orgs → `reset_events` + bootstrap; Stripe adaptive pricing → fix 4), re-verified green after remediation; one interim run hit the wedged backend now filed as BE #693 (113 passed / 0 failed / 152 self-skipped on the backend-down probe).
- `make check` green (format, lint `--max-warnings 0`, svelte-check 0 errors, i18n 4/4 locales, file-length, SSR-token, image-alt); unit tests 846 passed.

https://claude.ai/code/session_01CxmNyfoaDPenbP76Bx5frb

### DIFF
--- a/src/lib/components/attendees/AttendeeCardList.svelte
+++ b/src/lib/components/attendees/AttendeeCardList.svelte
@@ -95,6 +95,9 @@
 						type="button"
 						onclick={() => onEdit(rsvp)}
 						class="inline-flex flex-1 items-center justify-center gap-1 rounded-md bg-secondary px-3 py-2 text-sm font-medium text-secondary-foreground transition-colors hover:bg-secondary/80"
+						aria-label={m['attendeesAdmin.editRsvpAriaLabel']({
+							name: getUserDisplayName(rsvp.user)
+						})}
 					>
 						<Edit class="h-4 w-4" aria-hidden="true" />
 						{m['attendeesAdmin.actionEdit']()}
@@ -103,6 +106,9 @@
 						type="button"
 						onclick={() => onDelete(rsvp)}
 						class="inline-flex flex-1 items-center justify-center gap-1 rounded-md bg-destructive px-3 py-2 text-sm font-medium text-destructive-foreground transition-colors hover:bg-destructive/90"
+						aria-label={m['attendeesAdmin.deleteRsvpAriaLabel']({
+							name: getUserDisplayName(rsvp.user)
+						})}
 					>
 						<Trash2 class="h-4 w-4" aria-hidden="true" />
 						{m['attendeesAdmin.actionDelete']()}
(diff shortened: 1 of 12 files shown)


biagiodistefano:
Final confirmation full-suite run (fresh backend after the #693 wedge, `--workers=4`):

**267 passed, 2 skipped, 0 failed, 0 flaky — 2.8 min.**

That completes the gate matrix: new suites 48/48 on 3× repeat, full suite clean on both projects + all regression projects.

PR #637, opened 2026-07-13
fix(reactivity,e2e): fix Svelte Query tracking gaps and VIES probe flakiness

florimond-collette:
This PR fixes three Svelte 5 reactivity tracking bugs caused by TanStack Query (v6) proxy integration, and one unrelated Playwright test flakiness risk when external web services are unavailable.

These issues were introduced in PRs #626, #629, #630, and #632 (E2E test suites under issue #591).

## Reactivity Tracking Gaps: Svelte 5 & Svelte Query v6

Svelte Query v6 uses observer-level property tracking (`notifyOnChangeProps: 'tracked'` by default), backed by Svelte 5 reactive proxies. Svelte 5's `$derived` runes are evaluated lazily.

When query properties appear in short-circuiting conditional logic (like `a || b` or `{#if a} ... {:else if b}`), the properties on the right side are **not read on initial render** and therefore **not registered as tracked** by the Svelte Query proxy. When the query later resolves or fails, Svelte Query does not update the Svelte state signals for those untracked properties — leading to frozen loading spinners, blank error pages, or enabled buttons on failed queries.

Fixes 1–3 below address this pattern. Fix 4 is independent.

### 1. Stuck Loading Spinner in SubscriptionDrawer

*   **File**: `src/lib/components/members/SubscriptionDrawer.svelte`
*   **Commit**: [`a11933f`](https://github.com/florimond-collette/revel-frontend/commit/a11933f)
*   **Problem**: `subQuery.isLoading || !sub` short-circuits on mount, leaving `subQuery.data` untracked. The drawer gets permanently stuck in a loading state. `paymentsQuery.data`, in the `{:else}` block that is not rendered on mount, suffers from the same gap.
*   **Fix**: Introduce an `isLoading = $derived.by()` block that reads `subQuery.isLoading`, `subQuery.data`, `paymentsQuery.isLoading`, and `paymentsQuery.data` unconditionally, registering all four as tracked dependencies on mount. Also cast `subQuery.data as SubscriptionSchema` to satisfy strict TypeScript typechecking in the template.

### 2. Enabled Purchase Button on Query Error in SeriesPassCard

*   **File**: `src/lib/components/series-passes/SeriesPassCard.svelte`
*   **Commit**: [`ceabdbe`](https://github.com/florimond-collette/revel-frontend/commit/ceabdbe)
*   **Problem**: `quoteQuery.isLoading || !!quoteQuery.error` short-circuits on mount, leaving `error` untracked. If the query fails, the stale `null` value is read instead of the actual error, leaving the purchase button enabled.
*   **Fix**: Introduce an `isButtonDisabled = $derived.by()` block that reads both `isLoading` and `error` unconditionally, ensuring the error state is registered as a tracked dependency from mount.

### 3. Blank Page on Query Failure in Discount Codes Edit Page

*   **File**: `src/routes/(auth)/org/[slug]/admin/discount-codes/[code_id]/edit/+page.svelte`
*   **Commit**: [`1ee56b5`](https://github.com/florimond-collette/revel-frontend/commit/1ee56b5)
*   **Problem**: `{#if codeQuery.isLoading} ... {:else if codeQuery.error}` short-circuits on mount. On query failure, Svelte reads the stale `null` error, falls through all branches, and renders a completely blank page instead of the error alert.
*   **Fix**: Introduce a `queryStatus = $derived.by()` block that unconditionally reads `isLoading`, `error`, and `data` on mount, ensuring all three are tracked before any query transition can occur.

### 4. Playwright Crash Flakiness When VIES Is Down

*   **File**: `tests/e2e/journeys/j16-billing/vat-preview.spec.ts`
*   **Commit**: [`2d8277f`](https://github.com/florimond-collette/revel-frontend/commit/2d8277f) and [`f579705`](https://github.com/florimond-collette/revel-frontend/commit/f579705)
*   **Problem**: The VIES pre-flight probe calls `api.post(...)`, which throws a raw `ApiError` if VIES is down or the backend proxy fails. This crashes the Playwright runner with an unhandled exception instead of gracefully skipping the test.
*   **Fix**: Wrap the probe `api.post(...)` request in a `try/catch` block to intercept network/gateway errors. Following code review feedback, scoped the `try/catch` block strictly to the API request to prevent catching Playwright's `TestSkipError` thrown by the downstream `test.skip(probe?.vat_id_valid !== true)` check, invoking `test.skip(true, ...)` only when actual connection errors are caught.

## Verification

- **TypeScript**: `pnpm check` (`svelte-check`) passes with 0 errors in the modified files.
- **Formatting**: `pnpm prettier --write` applied to all changed files.
- **E2E**: `vat-preview.spec.ts` skips cleanly (without throwing) when the VIES endpoint is unavailable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved loading/error handling across the subscription drawer, series pass buy button, and discount-code edit page to better reflect current query state.
  * Ensured the UI updates correctly as payment-related data changes while subscription details are loading.

* **Tests**
  * Hardened VAT-ID journey tests to skip gracefully when the VAT preview service is unavailable or returns an invalid result, avoiding unrelated test failures.


biagiodistefano:
### Code review

Found 1 issue:

1. The new try/catch around the VIES probe also wraps the "not validated" `test.skip(...)` call. Playwright's `test.skip(condition, reason)` throws a `TestSkipError` when the condition is truthy, so when VIES responds but doesn't validate the ID, that skip throws *inside* the try and is caught by the new `catch (err)`, which re-skips with `VIES probe request failed: Test is skipped: VIES did not validate ...` — a doubly-wrapped, misleading reason claiming the request failed when it actually succeeded. This defeats the PR's own goal of distinguishing "service unavailable" from "won't validate the ID". Scope the try to just the `api.post(...)` call (keep the `probe.vat_id_valid !== true` skip outside it), or rethrow `TestSkipError` from the catch.

https://github.com/letsrevel/revel-frontend/blob/2d8277f1794e72b2d9a777510ced97aeafbce475/tests/e2e/journeys/j16-billing/vat-preview.spec.ts#L45-L64

florimond-collette:
Thanks for the catch @biagiodistefano! You're completely right. I've scoped the `try/catch` strictly to the API call so we don't intercept the skip error, committed the fix in `f579705`, and pushed it.

PR #646, opened 2026-07-14
feat(checkout): two-step online checkout — reserve, then create Stripe session (#464)

### LINKED ISSUE #464: feat(checkout): two-step online ticket checkout (reserve → create Stripe session) [blocked: BE #530]
**Blocked by** letsrevel/revel-backend#530 — do not start until the backend two-endpoint contract lands. Endpoint names/shapes below are provisional and owned by the backend issue.

## Context

The backend is splitting **online (Stripe) ticket checkout** into two requests so it stops holding a DB row lock across the Stripe API call (full rationale in backend #530). This changes the API contract the frontend consumes for **all online ticket purchases** — regular and PWYC, authenticated and guest.

## Current flow (single call)

`POST .../checkout` (or `/checkout/pwyc`) → `{ checkout_url }` → redirect to Stripe.

## New flow (two calls, provisional shapes)

1. `POST .../reserve` → reserves capacity + creates PENDING tickets → returns a `payment_id` / reservation handle. **Fast.**
2. `POST .../payment/{id}/session` → creates the Stripe Checkout Session → returns `{ checkout_url }`. **~1s.**
3. Redirect to `checkout_url`.

Perceived latency is unchanged (step 1 is fast; step 2 is the same ~1s Stripe wait we already pay), just two chained awaits behind the existing "Buy" action.

## Frontend work

- Regenerate the API client (`pnpm generate:api`) once the backend ships the new OpenAPI spec.
- Chain the two calls behind the existing purchase action.
- Loading/disabled state spanning both calls; guard against double-submit (step 1 now creates real PENDING rows server-side).
- **Partial-failure handling:** step 1 succeeds but step 2 fails. Surface a retryable error — the reservation auto-expires server-side, so retry is safe (no client-side cleanup needed).
- Cover all four entry points: authenticated regular + PWYC, guest regular + PWYC.

## Out of scope

- Free / offline / at-the-door tiers (no Stripe redirect) — unchanged.

biagiodistefano:
Closes #464

## What

Adapts every online (Stripe) purchase flow to the backend's reserve → checkout-session split (letsrevel/revel-backend#632, PR letsrevel/revel-backend#702): the checkout endpoints now only **reserve** capacity (returning `reservation_id` + `requires_payment`), and a second idempotent POST to the matching `checkout-session` endpoint returns the Stripe URL.

- **Shared helper** `src/lib/utils/checkout-session.ts` — routes `user` / `guest` / `series-pass` to the right session endpoint; typed `CheckoutSessionError` carries the `reservation_id` and an `expired` (404) flag.
- **Authed regular + PWYC** — session call chained inside the mutations, so the pending/disabled state spans both requests (double-submit guard). Partial failure (reserve ok, session fails) refreshes user status so the existing **Resume Payment** action appears — the backend recreates sessions for un-sessioned reservations, so that path is the retry.
- **Guest regular + PWYC, series passes** — the dialogs keep the reservation handle; a retry submit skips re-reserving and re-calls the idempotent session endpoint; on 404 (expired) they fall back to a fresh reserve automatically.
- Retryable payment-start-failed messages in en/de/fr/it; free/offline/at-the-door flows untouched.
- API client regenerated from the BE branch spec.

## Deploy ordering

**FE first, then BE.** The helper passes through a `checkout_url` already present on the reserve response, so this FE works against the current backend; the reverse order leaves the old FE silently no-oping on online purchases (`checkout_url` is `null` under the new contract).

## Testing

- `make check` green; unit suite **857/857** incl. 11 new helper tests (routing, 404-expired semantics, network failure, legacy passthrough).
- **34/34 affected e2e** against the BE #632 branch (j06 Stripe/batch/seats/self-cancel/door-PWYC, j16 VAT, j20 discounts, j22 invoicing ×2, j26 passes ×2) — full paid Stripe → webhook → ACTIVE journeys, desktop + mobile.
- **New** `j07-guest-attendee/guest-online-checkout.spec.ts` (guest online was previously uncovered): fixed-price full paid round-trip + PWYC redirect, 4/4 green. `startOnlineCheckout` factory migrated to the two-step contract.
- Full e2e suite green (the only failures were the known org-list pollution class in j01, fixed with a seed reset and re-verified).

https://claude.ai/code/session_01CxmNyfoaDPenbP76Bx5frb

## Summary by CodeRabbit

* **New Features**
  * Online checkout now reliably resumes previously held ticket or pass reservations when payment-session creation needs to be retried.
  * Guest, ticket, and series-pass purchases support the updated payment flow and redirect customers to hosted checkout.
* **Bug Fixes**
  * Added clearer, localized error messages when payment cannot be started.
  * Improved handling of expired reservations and interrupted checkout attempts.
* **Tests**
  * Expanded automated coverage for guest checkout and payment-session scenarios.


biagiodistefano:
### Code review

biagiodistefano:
Addressed the retry finding in 9eb82b24: a fingerprint-keyed retry store (`createReservationRetry` in `checkout-session.ts`) now backs both the authed controller and the guest dialog. An identical Buy-again replays only the idempotent checkout-session call instead of issuing a second reserve (which stranded the first reservation and could trip `max_tickets_per_user`, since PENDING tickets count toward the cap). A submit with changed parameters — or an expired reservation (404) — drops the handle and reserves afresh; this also fixes the guest dialog resuming a reservation priced with a stale PWYC amount/billing after the buyer edited the form. Covered by 6 new unit tests; affected e2e suites (40 tests incl. the guest Stripe round-trip) re-run green.

PR #665, opened 2026-07-19
fix(seating): close paint round-trip + sector kind prefill (BE #734)

biagiodistefano:
## Summary

Backend #734 exposes the seat's painted `price_category` (+`price_category_id`) and the sector's `kind` on the **admin read schemas** — the round-trip gap I filed as revel-backend#733. This wires the frontend consumers. **Targets `feature/venue-seating`** (ships with the coordinated seating deploy).

- **Grid-editor paint round-trip (the fix)**: existing seat paint now re-hydrates on reload. `readExistingPaint` already read `price_category_id` defensively (written anticipating #734), so the paint now shows — closing the "paint invisible after reload" gap the earlier browser smoke test flagged. Also cleaned the now-stale "backend follow-up" comments.
- **SectorModal kind prefill**: the edit modal prefills the sector's actual `kind` instead of the "Keep current type" placeholder; a seated-with-seats sector stays locked; the dead `kindKeepCurrent` key is retired across all four locales.
- **Sector page**: the standing empty-state reads the real `kind` (drops the defensive cast/fail-safe).
- Client regenerated for #734 (also carries the BE #731/#732 box-office surface already consumed; `kind` tightened to the `seated|standing` enum).

## Testing
- `make check`: **0 errors**; **1146 unit tests** pass; i18n 100% ×4.
- **Seating e2e green** against the live backend; `venues.spec.ts` J19.3 now asserts **paint survives a reload** (previously UI-feedback only, since the field was absent).
- **Browser smoke test (Opus, live backend)**: paint visibly survives reload (and unpaint persists); sector kind prefills correctly (seated locked-with-seats, new standing prefills "Standing", no dead option). No bugs, no console errors on the affected paths.

## Notes / deferred
- **Designer seat-category coloring** — #734 also makes it possible to color the designer canvas seats by their painted category (currently a passive legend). Deferred as an enhancement (the file is at 725/750 lines and it's polish, not a fix). Good follow-up.
- Doc clarification (BE 29596c3d): accessible-seat protection is unconditional (no exhaustion fallback) — the FE already surfaces the distinct 409s (`no_block` / accessible-exhausted copy), no change needed.

Closes revel-backend#733 (frontend half)


biagiodistefano:
Reopening from a feature/* branch so the CI + Security workflows actually run (they only trigger on push to feature/* or PRs to main/develop — a fix/* branch to the integration branch skips Build/Lint/Unit).

PR #672, opened 2026-07-20
feat(seating): per-seat-category pricing (#668) + refund_unmatched notification (#671) — BE #739 sync

biagiodistefano:
Syncs the frontend with backend `feature/739-category-pricing` (BE #739 per-seat-category pricing, #741/#744 refund_unmatched notification, #740/#743 fixes) — regenerated client + the two FE work items it unblocked.

## Per-seat-category pricing (#668)

**Buyer**
- Seat map & list append each seat's **server-resolved** price to its accessible name/tooltip — dumb lookup per the contract (`seat.price_category_id → seat_pricing.categories[id].price`, null → `unpainted`), never a locally recomputed fallback chain.
- Price legend in the picker (category swatch · name · price; unpriced categories flagged "Not available"), running **estimated total** for the held seats, and an honest min–max range on the tier card.
- Seats painted with a category the tier doesn't price render **blocked** (checkout would refuse them with a 400).
- **Staleness**: the hold controller refetches the chart whenever availability echoes a moved `chart_updated_at` — verified live by repainting a seat while the dialog was open (price flipped €20 → €55 after the next hold round-trip).
- Instant-issue paths (offline/at-the-door/free) now show the per-ticket amount on MyTicket **and** MyTicketModal ("Amount due:" while pending, "Price paid:" after) from `price_paid`.

**Admin**
- `TierFormSeatingSection` gains per-category price inputs for the selected sector, rehydrated from `TicketTierDetailSchema.category_prices`, with a coverage-gap warning from `pricing_gaps`.
- **Wipe protection**: `category_prices` follows the PUT three-way write semantics (omit = untouched, `{}` = clear, non-empty = replace) and is sent **only when dirty** — verified by intercepting a rename-only PUT (no `category_prices` key; server prices intact).

## refund_unmatched notification (#671)

- New notification type renders with a "Review tickets" CTA wired to `context.resolve_url` (BE #744) — verified live: injected notification → CTA → event ticket admin lands with `?search=<buyer email>` pre-filled; the table already shows seat + price paid so same-buyer tickets are distinguishable by amount.
- Preference toggles need no FE work (the type list is API-driven).

## Verification

- `make check` green (format, lint 0/0, svelte-check, i18n ×4 locales incl. updated paint explainer, file-length).
- `make test` green (full unit suite; new suites for `seat-pricing`, `buildCategoryPricesPayload`, category-blocked seat views).
[...]

Issue #668, opened 2026-07-20
Seating: per-seat-category pricing for user_choice seat selection (needs BE)


biagiodistefano:
## Context

Surfaced while smoke-testing seat **price-category painting** in the venue grid editor. Painting a seat sets its `default_price_category` (a label + color), and this works and persists correctly. The confusion is about what painting *does downstream*.

Today a painted category only affects **pricing** through **`best_available`** ticket tiers: a `best_available` tier is bound to exactly one price category (`TicketTier.price_category`) and auto-assigns seats painted with that category, charging the tier's single flat price. See `revel-backend/src/events/service/seating/pick.py` (loads "holdable seats in the tier's price category").

## The gap

**`user_choice` seating cannot price by seat category.** A `user_choice` tier is bound to a *sector*, not a category, and a `TicketTier` has one flat price. So when a buyer picks their own seat, every seat in the sector costs the same regardless of its painted category. On the buyer seat map painted seats get only a subtle 2px category-colored outline (`SeatMap.svelte`), no price.

Net effect for an admin: they paint Premium/Standard/Economy expecting three prices for hand-picked seats, sell via "pick your own seat", and see a prettily-outlined map at one price — i.e. *"painting has no effect."*

## Desired behavior

When seats are painted into price categories and the tier uses `user_choice`, the price a buyer pays should be derived from the **selected seat's price category**, so admins can offer per-seat-category pricing while letting buyers choose their exact seat (the standard theater model: pick your seat, price depends on where it is).

## Work involved

**Backend (the blocker):**
- Allow a price to be associated with a price category (per event or per tier), so a `user_choice` seat's price resolves from its `default_price_category` rather than the tier's flat price.
- Pricing/checkout + hold/reserve paths (`batch_ticket_service`, seating purchase) must compute the per-seat price from its category for `user_choice`.
- Public seating chart already exposes `price_category_id` + category colors — extend to expose the resolved **price per category** for buyer display.

**Frontend (after BE):**
- Tier form (`TierFormSeatingSection.svelte`): expose category pricing under `user_choice` (currently the price-category selector only appears for `best_available`).
- Buyer seat map (`SeatMap.svelte`): show the price per category (legend + selected-seat total), not just a color accent.

## Notes
- No functional bug in painting itself — this is a capability gap.
[...]

Issue #663, opened 2026-07-18
Seating phase 4 (box office): door sales, comps, reseat, seat on check-in (FE for BE #731)


biagiodistefano:
Frontend for backend #731 (seating phase 4 — box office), already merged on `feature/venue-seating` and in the regenerated client. All endpoints are event-admin, gated on `manage_tickets` — they belong alongside the box-office **overrides** page shipped in #658 (`/org/[slug]/admin/events/[event_id]/seating`).

### 1. Door sale / comp — `eventadminseatingBoxOfficeSell`
`POST /event-admin/{event_id}/seating/sell` — `BoxOfficeSellRequest {seat_id*, tier_id*, payment_method* (at_the_door|free), email? | user_id?, guest_name?, first_name?, last_name?}` → `AdminTicketSchema`.
- [ ] "Sell at the door" flow on the event-admin Seating surface: pick a seat (reuse the sector→row→seat list / SeatMap), pick a tier (seated tiers for that sector/category), choose recipient (email — guest get-or-create, with optional first/last/guest name — OR search an existing user), payment method (At the door / Comp=free).
- [ ] Result: success toast + issued ticket summary. Invalidate availability so the seat flips to sold.
- [ ] Errors: KILLED seat 400, foreign live hold 409, capacity/tier gates (reuse the purchase error surface) — localized.

### 2. Reseat — `eventadminseatingBoxOfficeReseat`
`POST /event-admin/{event_id}/seating/reseat` — `BoxOfficeReseatRequest {ticket_id*, target_seat_id*}` → `AdminTicketSchema`.
- [ ] "Move seat" action for an existing seated ticket: from the event tickets/attendees list (per-ticket action) or the Seating page — select the ticket, then pick a free target seat in the SAME price category; occupied/overridden/foreign-held targets rejected (localized).
- [ ] Invalidate availability + the ticket/attendee list; toast with the new seat.

### 3. Check-in seat display — MinimalSeatSchema in scan payload
The check-in scan response now carries `seat` (MinimalSeatSchema) + `sector_name`.
- [ ] CheckInDialog already renders `seat`; extend it to show `sector_name` + row/label ("Stalls · Row C, Seat 12") so door staff can direct attendees. Verify checkin-adapter maps the new fields.

## Cross-cutting
[...]

Issue #658, opened 2026-07-18
Seating: box-office seat overrides admin UI (Journey 19.7)


biagiodistefano:
Follow-up to #657 (seating phase 2). Backend Journey 19.7 (`USER_JOURNEYS.md`) ships a box-office seat-control API with **no frontend surface yet**:

`PUT /api/event-admin/{event_id}/seating/overrides` (permission: `manage_tickets`) — SDK op `eventadminseatingApplyOverrides`, already in the generated client.

## Scope

**Event-admin "Seat Control" surface** (event admin area, gated on `manage_tickets`):

- [ ] View current per-event overrides: seat, status (`held` / `killed`), reason. The chart endpoint (`eventpublicseatingGetChart`) + availability (`blocked` entries) provide the read side; overrides themselves have no dedicated list endpoint — derive from availability + chart, or request a BE read endpoint if insufficient (check with BE first).
- [ ] Bulk **set** overrides: pick seats (by sector → row/label multi-select; the phase-3 seat map can replace the picker later — do not block on it), choose `held` (house/tech/promoter) or `killed` (not sellable this event), free-text `reason`.
- [ ] Bulk **release** overrides by seat.
- [ ] **Per-seat rejection reporting**: the response returns `applied` / `released` counts and a per-seat `rejected` map (`"ticketed"` = seat has a non-cancelled ticket on this event, `"unknown_seat"`). Surface these per seat, never as a whole-batch failure. If a seat is in both set and release, release wins (backend rule — mirror in the UI copy).
- [ ] Effect feedback: overridden seats read as `blocked` in the public availability (buyers can't pick them; best-available skips them). Invalidate `['seating-availability', eventId]` after mutations.
- [ ] i18n (en/de/it/fr), WCAG AA (keyboard multi-select, focus management), mobile layout.
- [ ] Unit tests (mock SDK per house pattern) + e2e (arrange overrides via API, assert buyer-side `blocked` rendering + admin rejection reporting).

## Context / references
- Backend controller: `revel-backend/src/events/controllers/event_admin/seating.py`; service `events/service/seating/overrides.py`.
- Journey 19.7 for exact semantics.
- The buyer-side `blocked` rendering already exists (`seating-view.ts` from #657); only the admin surface is missing.

Part of the seating overhaul — target the `feature/venue-seating` integration branch.

biagiodistefano:
[...]

PR #673, opened 2026-07-20
feat(seating)!: venue seating — reserved seats, category pricing, designer, box office

biagiodistefano:
The complete venue-seating rollout (integration branch `feature/venue-seating` → main): 8 squashed PRs, 132 files, ~21k insertions.

> [!IMPORTANT]
> **Coordinated deploy — do not merge until the backend `feature/739-category-pricing` line is deployed.** This adopts the new seating API wholesale (BREAKING: legacy `random` seat mode removed, hold-based selection, category pricing schemas). Merge manually; auto-merge is deliberately not enabled.

## What's included

- **#661 (#657)** — Seating phase 1/2 API adoption: hold-based seat picker ("selection IS the server hold", TTL holds, 409 conflict surfaces), best-available tiers, price categories, guest seated flows.
- **#662 (#658/#659/#660)** — Phase-3 close-out: positional SVG seat map (buyer picker + pan/zoom + full a11y roving grid), venue designer, seat painting, box-office seat overrides, debt roll-up.
- **#664 (#663)** — Box office: door sales, comps, reseat, seat assignment at check-in (BE #731/#732).
- **#666 / #667** — Paint round-trip fixes, sector kind prefill, designer canvas seats colored by category (BE #734).
- **#669** — Sector-block venue designer with stage placement (BE #737), buyer stage-direction indicator for scoped sector views, drag/rotation fixes, admin seat-pricing UX (stale-cache eviction, discoverability).
- **#672 (#668/#671)** — Per-seat-category pricing for `user_choice` tiers (BE #739): server-resolved per-seat prices on the buyer map/list, price legend + estimated total, honest min–max tier price range, `chart_updated_at` staleness refetch, `price_paid` on tickets, tier-form category price editor with PUT wipe-protection; `refund_unmatched` notification with resolve deep link (BE #741/#744); precise paint-time coverage warning from the paint report (BE #746).

## Verification along the way

- Every PR landed gate-green (format, lint 0/0, svelte-check, i18n ×4 locales, file-length) with the full unit suite (now 1210 tests).
- All 27 seating-related e2e specs green (j06 tickets incl. the new category-pricing scenario, guest-seated, seating-self-service, venue-designer, venues, seat-overrides); seat-selection also on mobile-chrome.
- Live browser smoke tests against the demo stack at each step (admin painting/pricing flows, buyer picker, staleness, notification deep link).
- Branch-push CI green on the integration head (`6cd797d7`).

## Deploy notes

- Backend must be migrated first (`0102_tickettier_category_prices`, notifications `0023` for the `refund_unmatched` type, plus the earlier seating-phase migrations on the BE branch).
- Frontend client is generated from the BE branch's OpenAPI spec — deploying this FE against the old API breaks seated checkout.

Related backend: letsrevel/revel-backend#739, #741, #744, #746, #734, #737, #731/#732.


biagiodistefano:
## Backend contract change landing before release — one rename, one new capability

Two things on the seat-paint endpoint (`PUT /organization-admin/{slug}/venues/{venue_id}/seats/paint`) that affect the grid editor. Both are on `feature/venue-seating` via BE #748, i.e. **before** this integration PR reaches `main` — so it is a clean break, not a migration.

### 1. `under_covered_tiers` → `affected_tiers` (breaking)

BE #746 shipped `under_covered_tiers` a few days ago. BE #747 replaces it, because under-coverage turned out to be the *less* damaging half of the problem.

Painting a seat between two categories that are **both priced** — Premium €80 → Standard €30 — changes what buyers are charged for **every event at that venue**, instantly, with no error anywhere. Under-coverage at least fails closed (the seat stops selling, a buyer gets a 400). A repricing fails open: sales just continue at the wrong price. `under_covered_tiers` was empty for that case.

So under-coverage is now a sub-flag rather than the entry condition:

```json
{"painted": 1,
 "affected_tiers": [
   {"tier_id": "…", "tier_name": "Stalls",
    "event_id": "…", "event_name": "Gala Night", "event_status": "open",
    "price_changes": [{"seat_count": 1, "from_price": "80.00", "to_price": "30.00"}],
    "missing_categories": []}]}
```

**Migration:** rename the array. The `missing_categories` objects inside each entry are byte-identical to what you have. The old list is one expression:

```ts
affected_tiers.filter(t => t.missing_categories.length)
```

Notes for rendering:
- `to_price: null` means checkout now **refuses** that seat — that is what makes under-coverage a *transition* rather than a separate concept.
- `price_changes` is an array, not a single object: one paint writes one target category but can overwrite several source categories, so there is one `to_price` per tier and many possible `from_price`s. Sorted biggest-move-first, so you can lead with a headline.
- `price_changes` is **this paint's delta**; `missing_categories` is the tier's **current** gap. Both are correct, but please don't render them as one sentence — a pre-existing gap keeps a tier in the list even when this paint moved no money.
- Entries are per-tier. An event with three category-priced tiers on the painted sector appears three times; group client-side if you want "N events".
- A same-category repaint, and painting into a category priced at exactly the tier's flat price, both correctly report **nothing**. An advisory that fires when nothing changed is the failure we are fixing.

### 2. Dry-run preview (in progress, lands on the same PR)

Product decision: the report above arrives *after* the UPDATE commits, which for an event with live sales is too late — the money already moved.

So the endpoint is gaining a `preview` query parameter. `?preview=true` returns the **identical payload** and writes nothing, so the editor can ask "what would this do?" and show a confirmation before committing.

It is deliberately the same code path with the write as the only conditional step — a preview that could diverge from reality would be worse than none, since an admin would confirm one thing and get another. Validation (unknown seat → 404, foreign-venue category → 400) still fires in preview, so you cannot confirm a paint that would then fail.

Suggested UX, entirely your call: preview on every paint-into-category save, and gate on explicit confirmation only when `affected_tiers` contains an event with live sales — otherwise the confirmation becomes noise and gets click-through'd.

I will follow up here with the exact parameter name and the regenerated OpenAPI once it lands. Nothing to build against until then; flagging now so you don't build the advisory twice.

### Related
- letsrevel/revel-backend#748 — the PR carrying both
- letsrevel/revel-backend#747 / #746 — the two halves of the problem
- letsrevel/revel-frontend#674 — the frontend issue for the repricing advisory

biagiodistefano:
## Dry-run preview has landed — final contract

Follow-up to my note above. The parameter is confirmed and on `feature/venue-seating` via BE #748.

`PUT /organization-admin/{slug}/venues/{venue_id}/seats/paint` now accepts **`?preview=true`** (default `false`). Same method, same body, same response schema. It computes the full `SeatPaintResultSchema` (`painted`, `affected_tiers` with `price_changes` and `missing_categories`) and **writes nothing**: no seat changes category, and `updated_at` does not move — so the seat-chart version is untouched and open buyer views are not invalidated.

**The preview payload is byte-identical to what the real save returns for the same request**, including `missing_categories`, which reflects coverage *after* the paint would land — not the current gap. Backend tests assert full-payload equality in both directions, so you can render the confirmation dialog straight from the preview response and re-render the same component from the save response.

Worth knowing why that guarantee is real rather than aspirational: a naive dry run had a genuine bug here, caught while proving the equality. `missing_categories` was computed from a live read of the sector, which in the real path runs *after* the UPDATE. In a preview that would have reported the **pre-paint** coverage — so painting seats into an unpriced category would have previewed "no gap" and then opened one on save. Exactly the confirm-one-thing-get-another failure a preview exists to prevent. Coverage is now derived identically in both modes and cannot observe whether the write happened.

**Validation is unchanged:** previewing a paint with a seat outside the venue still 404s, a foreign-venue category still 400s, and nothing is written in either case. Treat a preview error as "this paint is not possible" and don't offer the confirm button. Permissions unchanged (`edit_organization`).

**Suggested flow** (your call): on save, call with `?preview=true`; if `affected_tiers` is empty — the common case by design — just save, no dialog. If non-empty, show the confirmation and, on confirm, repeat the identical request without the param. `painted` in the preview is how many seats *would* be painted.

⚠️ **Preview is advisory, not a reservation.** Nothing stops another admin repainting between the preview and the confirm, so the dialog can be stale. The save response is always the authoritative report — please render it, not just the preview. If that staleness ever matters in practice the fix is optimistic concurrency on the sector version, not holding a lock across a human's thinking time; not built, and not needed unless you see it.

Query cost, for reference: preview is 8 queries, the real save 9 (the difference is the UPDATE), both flat whether you paint 2 seats or 200.

I'll post the regenerated OpenAPI once #748 merges into `feature/venue-seating`.

PR #675, opened 2026-07-20
feat(seating): surface silent repricing when repainting between categories (#674)

### LINKED ISSUE #674: Paint advisory misses silent repricing when repainting between priced categories
## Context

The grid editor's paint advisory currently warns about **under-coverage**: painting a seat into a category a tier doesn't price makes those seats unsellable. BE #746 made that precise — the paint response now returns `under_covered_tiers`, so the warning can name the affected tiers and events, or stay silent when nothing is affected.

There is a second, quieter hazard the advisory does **not** cover, and it is the more damaging one.

## The gap

Repainting a seat **from one priced category to another** — Premium (€80) → Standard (€30) — changes what buyers are charged, immediately, for **every event at that venue**. Coverage stays complete, so nothing fires: no validation error, no checkout refusal, `under_covered_tiers: []`, no gap on the tier.

Under-coverage fails *closed* (the seat stops selling, someone notices). A repricing fails *open* — sales just continue at the wrong price.

Unpainting has the same shape: the seat silently drops to the tier's flat price.

## Why the frontend matters here

The admin's mental model is **"I painted a seat"** — a colour change on a map they are editing for one upcoming event. The actual semantics are **"I moved a seat between price zones for every event at this venue, effective now."** That gap between intent and effect is a UI problem before it is an API problem; the backend can only report facts.

## Work (blocked on BE #747)

BE #747 proposes extending the paint response from `under_covered_tiers` to `affected_tiers` — every category-priced tier whose sector was touched, with under-coverage as a sub-flag. Once that lands:

- Replace the single under-coverage advisory with two distinct messages: **"these seats are now unsellable on N tiers"** (existing, fails closed) and **"this changed prices for N events"** (new, fails open).
- Make the venue-wide blast radius visible *before* the paint commits, not only after — the editor is opened in the context of one event, which is exactly what makes the venue-wide effect surprising.
- Consider a confirmation step when a repaint would reprice seats on an event with live sales. An after-the-fact toast is probably not enough for that case.

## Not in scope

Blocking the paint. Painting is deliberately allowed to always succeed (BE #739 §4.3) so one event's pricing can never block another event's map work.

## Related

- letsrevel/revel-backend#747 — the backend half
- letsrevel/revel-backend#746 — under-coverage reporting
- letsrevel/revel-backend#739 — per-seat-category pricing

biagiodistefano:
Implements #674 in full — both the FE-only slice and, now that BE #747 landed (`affected_tiers` + `?preview=true` dry run, BREAKING: replaces `under_covered_tiers`), the precise half.

**The hazard**: repainting a seat between priced categories (or unpainting it) reprices it for **every event at the venue**, immediately, and fails open — coverage stays complete so nothing else fires.

**What ships:**
- **Pre-commit blast radius**: the paint palette carries a caution that paint is venue-wide and immediate.
- **Confirmation on live money**: saves with paint batches dry-run the paint (`?preview=true`) *before any mutation*; a price change on an **on-sale (open)** event raises a confirm naming the tier, event, and exact move ("1 seat(s) 80.00 → 30.00"). Cancel aborts the whole save — nothing persisted. Drafts and coverage gaps (fail closed) never prompt.
- **Precise post-paint report**: two distinct toasts from the real response — the repricing summary (the previously-silent fails-open case) and the under-coverage warning (fails closed, with the "Fix tier prices" deep link). `mergeAffectedTiers` dedupes per tier across batches.
- Client regenerated for the new response shape; the interim client-side reprice heuristic from the first commit is replaced entirely by server truth.

**Verified**:
- Live browser (the exact silent scenario — Front €55 → Balcony €30 with both categories priced on an open event): confirm shows the precise move; cancel → server state untouched; accept → paint applied both directions.
- New e2e in `venues.spec.ts` drives the full flow — palette caution, repaint, confirm dialog assertions (ON SALE, tier, event, move), accept, "Repainted seats changed price zone" toast with the precise line. All venues + seat-selection specs green.
- Gate green with real exit codes (1211 unit tests incl. `mergeAffectedTiers`/`liveRepricedTiers`), i18n ×4 locales.

Closes #674


biagiodistefano:
BE #747 landed mid-review, so this PR now carries the FULL #674 scope: preview-based confirmation for live-event repricing (cancel aborts the save cleanly), precise per-tier/per-move toasts from `affected_tiers`, and the client-side heuristic replaced by server truth. New e2e covers the confirm dialog + precise toast end-to-end. Body updated; now `Closes #674`.

Issue #704, opened 2026-07-28
Type gate is vacuous: svelte-check 4.7.3 emits no TypeScript diagnostics under typescript ^6.0.3


biagiodistefano:
## Summary

`make types` — and the CI job that runs it — has been passing **vacuously**. `svelte-check` 4.7.3 against the pinned `typescript` ^6.0.3 does not fail, and does not warn: it silently emits **zero TypeScript diagnostics**. Every `.svelte` file in the repo currently has **no type coverage at all**, and CI reports green regardless of what they contain.

This is distinct from #645. That issue tracks adopting TypeScript **7**, and notes that `svelte-check` *crashes* against it (`TypeError: … reading 'Cjs'`) — a loud, obvious failure, and it implies TS 6 is the working state. It isn't. On TS 6 the failure is silent, which is worse: a crash stops the build, a vacuous pass ships.

## Reproduction (~2 minutes)

1. Append a deliberate type error to any `.ts` file, e.g. in `src/lib/utils/format.ts`:
   ```ts
   export const __probe: number = 'definitely not a number';
   ```
2. `pnpm svelte-check --tsconfig ./tsconfig.json` → **0 errors**. Not caught.
3. `npx tsc -p tsconfig.json --noEmit` → caught, `TS2322`.
4. Revert the probe.

## Do not use the FILES count as a health signal

It drifts wildly for the same tree — 6042, 563, 204, 99 observed across runs and environments — because it reflects whatever `svelte-check` managed to load, not the project.

Likewise, a run reporting "0 errors, N warnings" proves nothing. Those warnings are Svelte **compiler** warnings (`state_referenced_locally` etc., the ones #361 tracks), not TS diagnostics. That output shape is exactly what a broken run looks like. Four separate parties each read a different files count as evidence the gate was fine; only the injected-error probe is decisive.

## Impact

- `.svelte` files: **no type checking whatsoever**, including props, bindings, and store/rune usage.
- CI's type job is green by construction and cannot fail.
- `.ts` files are still checkable via `tsc` directly, but nothing in the gate does so.

Concretely: several branches' worth of `.svelte` changes have been merged with the type gate disarmed. No known bug has been traced to this, but that's not evidence of absence — it's the absence of a detector.

## Interim check

```bash
npx tsc -p tsconfig.json --noEmit
```

[...]

Issue #690, opened 2026-07-27
feat(events): visibility toggles UX + null-tolerant rendering for hidden capacity/counts/guest list


biagiodistefano:
Frontend counterpart of letsrevel/revel-backend#792 (granular event visibility settings). Blocked on the backend landing + `pnpm generate:api`.

## Feature

Organizers can hide organizational info per event: attendee count, capacity/spots-left, and the guest list (event-level override that ANDs with per-user prefs — can only hide, never reveal). Backend adds `visibility_settings` to the event schemas and an always-public `is_full` boolean; hidden fields will be absent/null in API responses.

## Organizer UX (event wizard/editor)

- New visibility section with granular toggles: `show_attendee_count`, `show_capacity`, `show_attendee_list` (alongside the existing address-visibility select and pronoun-distribution checkbox).
- **All / none / some** affordance: preset shortcuts (e.g. "Open" = everything visible, "Discreet" = counts + guest list hidden) that just set the granular toggles — presets are FE-only, backend stores granular values.
- Round-trip `visibility_settings` in `event-payload.ts`, `EventWizard.svelte`, `EventEditor.svelte`, and the series/template editors (`TemplateEditDialog.svelte`, `RecurringEventWizard.svelte`).

## Attendee-facing rendering — null-tolerance fixes needed

Most components already degrade gracefully, but these break when counts become absent:

- [ ] `EventDetails.svelte` (~L28-47): capacity guards use `=== undefined` — a `null` `max_attendees` passes the guard and renders "Full" (`null - count` < 0). Switch to `== null`-style checks.
- [ ] `EventRSVP.svelte` (~L176-179): optimistic `attendee_count ± 1` produces `NaN` when the field is absent — guard it.
- [ ] `EventQuickInfo.svelte` (~L130-149): `attendee_count` interpolated unguarded → "undefined / 50 spots".
- [ ] Sold-out badges (`EventBadges.svelte`, `EventStatusBadge.svelte`, `utils/event.ts::isEventFull`): currently derive fullness from `max_attendees`/`attendee_count` with `?? 0`, so hidden capacity silently reads as "not full" — switch to the new backend `is_full` boolean.
- [ ] `seo/jsonld/event.ts` (~L82-84): sold-out `eventStatus` derivation — use `is_full`.
[...]

Issue #653, opened 2026-07-16
Tech debt: generated 400 response types are unreliable (backend returns 3 body shapes)


biagiodistefano:
## Summary

Known technical debt — frontend counterpart of letsrevel/revel-backend#712.

The generated API client types each endpoint's 400 response as the single schema declared in the backend's OpenAPI spec, but the backend actually returns **three different 400 body shapes** depending on which code path rejected the request:

- `{"detail": "..."}` — raised `HttpError` / app exception handlers
- `{"errors": {field: [msgs]}}` — Django model validation
- `EventUserEligibility` — eligibility rejections

Per the backend audit, ~17 endpoints can return a 400 body that does not match the generated type. Highlights that affect the frontend directly:

- **`ticket_checkout` / `ticket_pwyc_checkout` / `rsvp_event`**: typed as `EventUserEligibility`, but capacity, seat-resolution, discount-code, Stripe-config, and RSVP-note errors all arrive as `{"detail"}` — so `error.reason_code` / `error.next_step` are `undefined` on those paths.
- **9 endpoints typed as `ResponseMessage` (`{message}`)** (waitlist join/leave, questionnaire submit, guest RSVP/checkout endpoints, confirm guest action, cancel checkout): **no reachable 400 ever has a `message` key** — real bodies are `{detail}` or the eligibility payload.
- **Admin endpoints typed as `ValidationErrorResponse` (`{errors}`)**: several can also return `{detail}` (series-pass coverage, check-in, questionnaire create/evaluate, template update).

## Impact

Any error-handling code that trusts the generated 400 type (e.g. reading `.message`, `.reason_code`, or `.errors`) silently gets `undefined` on mismatched paths and may render blank error states.

## Follow-up

Harmonization will land on the backend first (see letsrevel/revel-backend#712 for options: honest unions, standardizing on `{"detail"}` + `reason_code`, or routing more errors through eligibility). After that lands: regenerate the client (`pnpm generate:api`) and sweep 400-handling call sites. Until then, treat 400 bodies as `unknown` and probe for `detail` / `errors` / `reason_code` defensively where it matters.

biagiodistefano:
Backend now localizes model-validation errors per Accept-Language (BE #724 line). Audit of FE English string-matching on backend messages (2026-07-21):

[...]

PR #706, opened 2026-07-29
fix: regenerate API client and null-tolerate the newly-nullable event/tier numbers

biagiodistefano:
Wave 0 of the frontend sweep against four merged backend PRs. **Mechanical correctness only** — no feature work, no new organizer toggles, no copy redesign. Three follow-up branches build on this one.

## What regenerated

`src/lib/api/` was regenerated against backend `main` (letsrevel/revel-backend#822, #824, #825) in the base commit `9e6fc1fa`. This PR is everything the app needed to be correct again afterwards.

Two shape changes matter:

- **#825 (granular event visibility)** — nine response fields went `int` → `int | null`: `attendee_count`, `max_attendees`, `seats_held` (event schemas), `total_available` (`TicketTierSchema`), `total_attendees` / `total_with_pronouns` / `total_without_pronouns` (`EventPronounDistributionSchema`), and `attendee_count` on both the dietary-restriction and dietary-preference aggregate entries. A new `visibility_settings` object and an always-public `is_full` boolean also appeared.
- **#824 (honest error schemas)** — many endpoints moved from `ResponseMessage` (`{message}`) to `ErrorDetail` (`{detail}`) or a union. Runtime behaviour never changed; the declarations had been lying, so `.message` was already `undefined` on those paths.

## The 29 type errors

Fixed by **suppressing the number, never faking one** — no `?? 0`, no `!`. `?? 0` is precisely what would make a withheld capacity silently read as "not full".

| Surface | Fix |
|---|---|
| `utils/event.ts` | `isEventFull` / `getSpotsRemaining` answer `false` / `null` when either number is withheld — fullness is not assertable without both |
| `EventDetails.svelte` | **Real bug**: the capacity guards tested `=== undefined`, so a `null` `max_attendees` passed and `null - count` went negative → the card rendered **"Event is full"** for every event that hides capacity. Now `== null` on both numbers |
| `EventQuickInfo.svelte` | "X / Y spots taken" needs both numbers; falls back to the raw count only when that is disclosed |
| `EventStatusBadge.svelte` | no "Full" badge without both numbers — falls through to the temporal states |
| `EventRSVP.svelte` | the optimistic `attendee_count ± 1` produced `NaN` against a withheld count; it now leaves the count withheld |
| `OccurrenceRow.svelte`, org-admin tickets list | omit the tally rather than render a fabricated `0` |
| `seo/jsonld/event.ts` | unknown occupancy stays `InStock`; structured data never publishes a guessed `SoldOut` |
| questionnaire summary + `AttendeeList` pronoun section | render the **existing "no data" state** when the totals are withheld. The backend serves the distribution *empty* in that case, so a zeroed chart would be fabricated. Also removed four pre-existing `?? 0`s that were papering over the same thing |
| `DietarySummary.svelte` | entries and notes still render; only the per-entry number is suppressed |
| `AttendeeList.svelte` | `totalAttendees` accepts nullish — the count chip and the "+N more" affordances disappear, the list itself still renders (it is gated separately by `show_attendee_list`) |
| org membership-questionnaire submit | reads `detail` instead of `message`, narrowed with `'detail' in err` **and** a `typeof === 'string'` check, because a ninja 422 `detail` is a *list*, not a string |

## ⚠️ Hazard A — `total_available: null` means two things

Before #825 `null` meant *unlimited*. It now also means *not disclosed*, distinguishable only via the event's `visibility_settings.show_capacity`. This compiles fine and would have been silently wrong.

**Finding, contrary to the brief:** `m['tierCard.unlimited']()` was already **unreachable in the rendered output**. `TierCard.svelte` computed it for a `null` `total_available`, but the template row was guarded by `{#if tier.total_available !== null && effectiveEligible}` — so the "Unlimited" string was never displayed, on the live site or anywhere else. I verified this is the only producer of that key (`grep -rn "tierCard.unlimited" src/ messages/` → one hit).

So there was no user-visible lie to fix, only a latent one. `availabilityStatus` now returns `message: null` for a `null` `total_available` (with the ambiguity documented), and the template row is driven by `availabilityStatus.message !== null` — **exactly equivalent to the old guard, so rendering is byte-for-byte unchanged**, and the "Unlimited" claim can no longer leak if anyone loosens that guard later.

**I did not add the `capacityDisclosed` prop** the brief asked for, and I want that called out rather than buried:
- It would have been **inert**. No rendering decision in `TierCard` depends on it: for `total_available === null` the inventory row is suppressed in *both* readings, and `availabilityStatus.available` is `true` in both (correctly — an undisclosed inventory is not evidence of sold-out, and the server enforces the real ceiling).
- Making it load-bearing would have meant *starting* to render "Unlimited" for disclosed-unlimited tiers, which is a UX change this PR is scoped against.
- Two of the three named mount sites were also wrong: `events/admin/TicketingStep.svelte` imports `./TierCard.svelte`, i.e. `events/admin/TierCard.svelte`, a different component. The real mount sites of `tickets/TierCard.svelte` are `TicketTierList.svelte` and `TicketTierModal.svelte` only.
- `events/admin/TierCard.svelte` reads `total_available ?? 0`, but it is staff-only surface and org owners/staff bypass all visibility gating, so it always receives real numbers.

Audited every other `total_available` reader (`grep -rn "total_available" src/ | grep -v src/lib/api/`). `TicketConfirmationDialog.svelte`, `GuestTicketDialog.svelte` and `venue-overview.ts` all already guard with `!== null` / `!= null`, so "only N remaining" and the max-selectable cap correctly say nothing for a withheld inventory. I documented the ambiguity at the two quantity-derived readers (`tierMaxSelectable`, `TicketTierModal.getTierPurchaseStatus`) so the next reader does not have to re-derive why `null` yields no client-side ceiling.

**Deliberate residue:** the `tierCard.unlimited` key is now unused in all four catalogs. Left in place rather than deleted — #690 is likely to want a capacity-disclosure string, and removing it would churn four files for no benefit.

## ⚠️ Hazard B — waitlist join 400 → 200

`POST /events/{id}/waitlist/join` now returns **200** on a double-submit / lost race where it previously returned 400.

Verified the existing reading is correct: `IneligibilityActionButton.svelte` special-cases 409 and routes all other errors to `showError`, so a 200 lands in the `else` success branch and self-heals. `grep -rn "JoinWaitlist" src/` confirms this is the **only** join-waitlist call site.

**No copy change needed** — and I want to correct the premise here too. The success message is `ineligibilityActionButton.success` = **"Success!"**, not "you joined". That is already honest for someone who was already on the waitlist, and the handler reloads 1.5 s later so the eligibility state re-renders as "on the waitlist" regardless. Changing generic copy would have been churn. (Separately: that key is untranslated in `it.json` and `de.json` — pre-existing i18n debt, not touched here.)

## Null-tolerance shape

Every guard treats an **absent/undefined `visibility_settings` as fully disclosed** — `visibility_settings?.show_capacity === false`, never `!visibility_settings?.show_capacity`, which would read an absent object as "hide everything". For the nine numeric fields the rule is the plain one: a missing number means *do not state a number*, so `== null` covers `null` and `undefined` alike.

## `is_full` deliberately not adopted

Sold-out badge derivation is untouched except where it was a type error. Adopting the new `is_full` boolean is feature work belonging to #690 and is handled on a separate branch.

## Small build hardening

`scripts/generate-api-client.js` resolved the spec as `resolve(__dirname, '../../revel-backend/.artifacts/openapi.json')`, which only lands correctly from the primary checkout. From any of this repo's 8+ git worktrees the file is absent and the run **silently** fell back to `BACKEND_URL || 'http://localhost:8000'` — quietly generating a client from whichever backend happened to answer. It now walks upward to find the real artifact (verified: resolves to `/Users/biagio/repos/letsrevel/revel-backend/.artifacts/openapi.json` from inside a worktree), accepts a `BACKEND_SPEC` override, and prints a loud multi-line warning when it does fall back.

## Tests

21 new tests, covering the withheld-number branches rather than just restoring green:

- `src/lib/utils/event.test.ts` (new) — `isEventFull` / `getSpotsRemaining` across disclosed, unlimited and withheld.
- `EventDetails.test.ts` — including an explicit regression guard for the `=== undefined` → "Event is full" bug.
- `EventQuickInfo.test.ts`, `EventStatusBadge.test.ts`, `seo/jsonld/event.test.ts`.

## Verification

Run locally on this branch, output observed:

- `make fix` — clean
- `make check` — **`COMPLETED 6053 FILES 0 ERRORS`**, plus `✓ Type gate is armed (deliberate errors in .ts and .svelte both caught)`. The canary matters: a green `make types` is meaningless if the gate has disarmed itself (#704). Warning count is unchanged from the pre-change baseline (373, all pre-existing `state_referenced_locally`).
- `pnpm test` — **163 files, 1965 passed** (baseline before this PR: 162 / 1944).
- `svelte-autofixer` — clean on every component touched. The one suggestion it raised (`priceDisplay` in `TierCard.svelte` passing a function to `$derived`) is pre-existing code this PR does not touch.

`pnpm test:e2e` not run, per instructions.

Refs #690, #653.


biagiodistefano:
@coderabbitai review

biagiodistefano:
@coderabbitai full review

PR #708, opened 2026-07-29
fix(errors): narrow the backend's honest error unions instead of casting (#653)

### LINKED ISSUE #653: Tech debt: generated 400 response types are unreliable (backend returns 3 body shapes)

biagiodistefano:
Backend PR [letsrevel/revel-backend#824](https://github.com/letsrevel/revel-backend/pull/824) made the OpenAPI error declarations honest, so most 4xx bodies are now `ErrorDetail`, `ValidationErrorResponse`, or an `anyOf` union of the two.

The regenerated client (Wave 0, #706) surfaced exactly **one** compile error — the app was already probing `'detail' in err` almost everywhere. But probing is not narrowing, and several sites then read the union through a cast that named a field the branch did not have. **This PR is the runtime half the type gate cannot see.** A green `pnpm check` was never evidence the work was done.

Closes #653

## Narrowing core — `src/lib/utils/api-error-detail.ts`

Real type predicates, no casts, no `any`:

| Guard | Shape |
|---|---|
| `isErrorDetail` | `{ detail: string }` — django-ninja domain refusals |
| `isValidationErrorResponse` | `{ errors: { field: string \| string[] } }` — model validation |
| `isResponseMessage` | `{ message: string }` |
| `isRequestValidationError` | `{ detail: [{ msg, loc, type }, …] }` — the **422 LIST** |

The 422 is the sharp edge called out in the issue: `detail` is a *list of objects*, not a string, and it is **not** `ErrorDetail`. `String(err.detail)` on one rendered `[object Object]`; `err.detail.toLowerCase()` would have thrown. 422 is systemically under-declared (backend #826), so it can arrive from an endpoint that never declares it — every reader now handles it defensively.

`backendMessage` now probes **detail (string or 422 list) → errors → message**. `message` moved *last*: outside the two claim-invitation endpoints its declaration was a lie and it was already `undefined` at runtime.

15 unit tests cover the guards; `backendMessage`'s precedence flip is asserted explicitly.

## Eligibility unions — the CTA re-render is the thing that must not regress

`rsvp_event`, `ticket_checkout`, `ticket_pwyc_checkout` are now `EventUserEligibility | ErrorDetail`.

- **`EventRSVP.svelte`** probes the eligibility branch first, keeps the payload as `Error.cause`, and **re-seats it into `userStatus` on error** — so a refusal re-renders the CTA against the *current* `next_step` instead of the stale one the page loaded with. Previously the refusal body was discarded entirely and the user saw the untranslated `"No data returned from RSVP"`.
- **`isNoteRejection`** now excludes eligibility refusals. It only checked `!('allowed' in error)`, so a refusal shaped slightly differently could be read as a note rejection and **silently re-submit an RSVP the backend had just refused**.
- **`event-checkout-controller.svelte.ts`** keeps reading the refusal first (preserving `isMembershipTierRefusal`) and falls through to the list-safe detail reader.
- `IneligibilityMessage.svelte`, `EligibilityStatusDisplay.svelte` and `utils/eligibility.ts` needed **no changes** — they consume a typed `EventUserEligibility` and never touch error bodies. Verified, not assumed.

4 new component tests on `EventRSVP` pin the behaviour: eligibility-400 adopts and re-renders the CTA; detail-400 shows the sentence and leaves the buttons; 422-list renders the messages and never `[object Object]`; a refusal triggers exactly **one** request.

## Waitlist join — the 400 → 200 behaviour change

The idempotent double-submit now lands in the success branch, which printed a generic `"Success!"`. It now prefers the backend's own sentence (*"You are already on the waitlist for this event."*), so the copy is honest for someone who was already on it. Its 400 probes eligibility → detail; the 409 special-case is untouched.

## Errors the backend now names, that were being thrown away

| Endpoint | Before | Now |
|---|---|---|
| `POST /event-admin/{id}/invitations` | unhandled 500 → blanket `fail(500)` | 400 `{detail}` **naming the offending tier ids**, real status |
| `POST /event-admin/{id}/tokens`, `PUT …/tokens/{id}` | hardcoded `'Failed to create token'` | newly declared 404, body names the offending tier ids |
| `POST /questionnaires/…/submissions/{id}/evaluate` | body discarded behind "Please try again" | declared 404 / 400 union surfaced, real status |
| `POST /events/claim-invitation/{token}` + org twin | read `detail` only → reason always dropped | the **only two** endpoints where `{message}` is genuine; all 4 call sites now read it |
| `POST /organizations/` (create-org) | `(error as {status?}).status` — never on the body, so both branches dead → always 500 | reads `response.status`; 400/403 reported correctly |

## Cast sweep

19 sites replaced `String(response.error.detail)` / `(response.error.detail as string)` with `extractApiErrorDetail`. `SubscriptionsTab`, `PollStatusBar`, `OrgContactEmailModal`, `NotificationPreferencesForm` and `StaffReviveModal` moved to `backendMessage` (the `{errors}` branch was being dropped). `PollStatusBar` no longer falls back to `JSON.stringify(error)` in a user-facing toast.

`CancelTicketDialog` validates the 409 `code` against the `CancellationBlockReason` enum instead of casting — that endpoint's 403/502 are among the twelve unaudited declarations.

## What I did NOT audit

- **The twelve remaining `ResponseMessage` error declarations** (backend #826): ticket cancellation-preview/cancel 403+502, the three checkout-session 404s, the two token-detail 404s, tier-seats 404, checkout-resume 404. Their declared types are **not trusted** — every reader treats the body as `unknown` and probes. No code assumes those declarations are accurate.
- **403/422 under-declaration.** Nothing here assumes an endpoint *cannot* return a status just because it is undeclared.
- **`guestAttendance.getLocalizedError`** — left as-is per the issue's own audit (graceful degradation: falls through to the backend message verbatim, which now arrives in the UI language via `Accept-Language`).
- **`RevivalRequestSchema.amount >= 0`** — already mirrored. `StaffReviveModal` is the only revive amount input and already has `min="0" step="0.01" required`; `RejoinCard` (online revive) sends an empty body. No change needed.
- **Plan-archive 502** — no 502 branch exists in the frontend, so there was no dead code to remove.
- **No new English string matcher was added.** `isBillingInfoRequiredError` already covers en/it/de/fr; everything else in this PR uses structured probing.
- **`/embed` + `/oembed`** (merged #707) reviewed at the coordinator's request: both use `{ data }` null-checks and log the error object structurally. Neither reads `.message` nor treats `detail` as a string. Nothing to fix.

## Verification

```
make check   # 0 ERRORS / 6082 files; "✓ Type gate is armed (deliberate errors in .ts and .svelte both caught)"
pnpm test    # 169 files, 2049 passed | 1 todo
```

Rebased onto `main` after #707 (`4ca8934a`) merged. E2E not run.

### DIFF
--- a/src/lib/components/events/ClaimInvitationButton.svelte
+++ b/src/lib/components/events/ClaimInvitationButton.svelte
@@ -8,6 +8,7 @@
 	import { goto } from '$app/navigation';
 	import { resolve } from '$app/paths';
 	import type { EventTokenSchema } from '$lib/api/generated/types.gen';
+	import { backendMessage } from '$lib/utils/api-error-detail';
 
 	interface Props {
 		tokenId: string;
@@ -42,11 +43,13 @@
 			});
 
 			if (response.error) {
-				const err: unknown = response.error;
-				const errorDetail =
-					typeof err === 'object' && err !== null && 'detail' in err ? err.detail : undefined;
+				// One of only two endpoints in the codebase that genuinely answer 400
+				// with `ResponseMessage` (`{message}`) — everywhere else that
+				// declaration was a lie and `.message` was undefined at runtime
+				// (backend #824). Its 404 is `{detail}`, so `backendMessage` probes
+				// both rather than picking one and dropping the other.
 				throw new Error(
-					typeof errorDetail === 'string' ? errorDetail : m['claimInvitationButton.failedToClaim']()
+					backendMessage(response.error) ?? m['claimInvitationButton.failedToClaim']()
 				);
 			}
 
--- a/src/lib/components/events/EventRSVP.svelte
+++ b/src/lib/components/events/EventRSVP.svelte
@@ -3,7 +3,14 @@
 	import { createMutation, useQueryClient } from '@tanstack/svelte-query';
 	import { eventpublicattendanceRsvpEvent } from '$lib/api/generated/sdk.gen';
 	import type { UserEventStatus } from '$lib/utils/eligibility';
-	import { isRSVP, isEligibility, isUserStatusResponse } from '$lib/utils/eligibility';
+	import {
+		isRSVP,
+		isEligibility,
+		isUserStatusResponse,
+		isEligibilityRefusal,
+		getEligibilityRefusalMessage
+	} from '$lib/utils/eligibility';
+	import { extractApiErrorDetail, isErrorDetail } from '$lib/utils/api-error-detail';
 	import { cn } from '$lib/utils/cn';
 	import RSVPButtons from './RSVPButtons.svelte';
 	import RsvpNoteDialog from './RsvpNoteDialog.svelte';
@@ -90,26 +97,15 @@
 	});
 
 	/**
-	 * The RSVP endpoint has two 400 shapes: the EventUserEligibility object
-	 * (carries `allowed`) and a plain `{detail}` when a note is sent while the
-	 * event no longer accepts notes. Detect the latter by shape, not wording.
+	 * The RSVP endpoint declares `EventUserEligibility | ErrorDetail` at 400
+	 * (backend #824). The eligibility branch carries `allowed`/`reason_code`; the
+	 * `{detail}` branch is what a note sent to an event that no longer accepts
+	 * notes produces (alongside capacity and RSVP-note rejections). Detect the
+	 * latter by shape, not wording — and never mistake an eligibility refusal for
+	 * it, which would silently re-submit the RSVP the backend just refused.
 	 */
 	function isNoteRejection(error: unknown): boolean {
-		return (
-			typeof error === 'object' &&
-			error !== null &&
-			'detail' in error &&
-			typeof (error as { detail: unknown }).detail === 'string' &&
-			!('allowed' in error)
-		);
-	}
-
-	function extractErrorDetail(error: unknown): string | null {
-		if (typeof error === 'object' && error !== null && 'detail' in error) {
-			const detail = (error as { detail: unknown }).detail;
-			if (typeof detail === 'string') return detail;
-		}
-		return null;
+		return isErrorDetail(error) && !isEligibilityRefusal(error);
 	}
 
 	// Mutation: RSVP to event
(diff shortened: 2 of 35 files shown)


biagiodistefano:
Both Copilot findings accepted and fixed in `0e327377`.

1. **`isRequestValidationError` unsoundness** — correct, and squarely the point of this PR. The predicate promised `RequestValidationItem[]` while only requiring *some* entry to carry a `msg`, so a mixed array narrowed to a type it did not have. It now requires a non-empty array whose *every* entry is a readable item; `extractApiErrorDetail` consequently drops its defensive re-filter and just maps. Two new tests pin the mixed-array rejection.

2. **`create-org` reporting non-400/403 as 500** — correct. That endpoint declares no error responses at all, so a runtime 422 is plausible (422 is systemically under-declared, backend #826) and must not read to the user as a server failure. Any 4xx now passes through, matching what the invitations and questionnaire-evaluate actions in this PR already do.

`make check` 0 errors (type gate armed), `pnpm test` 2050 passed.

biagiodistefano:
Third finding (`isValidationErrorResponse` unsoundness) also accepted — fixed in the latest commit. Same class as the 422 one: the guard checked only that *some* value was readable, so a map with a non-`string | string[]` value narrowed to a type it did not have. Every value is now validated against the declared shape; at least one readable message is still required so an all-blank map falls through to the caller's own localized copy rather than rendering as an empty string. Three new tests.

PR #703, opened 2026-07-28
feat: membership subscriptions — tiers, plans, Stripe billing, applications

biagiodistefano:
## Summary

The complete FE side of the membership-subscriptions launch, developed on this integration branch across five staged PRs plus follow-up waves:

- **PR① – ④**: org membership tiers & subscription plans (admin CRUD, per-tier questionnaires, approval overrides), hosted Stripe Checkout subscribe/manage/cancel/change-plan flows, member-facing membership hub, requests/settings admin surfaces, subscription metrics.
- **PR⑤ (#692)**: full Playwright E2E suite for journeys J23 (subscriptions) and J27 (applications), including the post-checkout `?membership_success` param-strip and CheckoutReturnCard polling coverage.
- **Follow-up waves**: FE #693–#697 (webhook write-back race cache seeding, abandoned-revival resume-payment action, status-breakdown metrics strip, dialog overflow sweep, polish batch), BE #774 review-fix adoption (`MembershipPaymentSchema` rename, partial-refund audit display, force-approve confirm flow, apply-endpoint eligibility-shaped 400 handling), BE #803 revival click-leg E2E, BE #804 tier-delete 409 surfacing.

## Test evidence

- Unit: **1636 passed / 0 failed** (150 files)
- E2E: **490 passed / 3 skipped / 0 failed, zero flakes**, validated twice — including once on a from-scratch stack (fresh Postgres/PgBouncer, canonical reseed, live Stripe webhook forwarder)
- `make check` green (lint `--max-warnings 0`, svelte-check 0 errors, i18n ×4 locales aligned)

## Deploy coordination

⚠️ **Do not merge until the coordinated FE+BE deploy**: requires `revel-backend` `feature/subscriptions-integration` deployed together (BE #776–#779 + #774 fix set + #794/#795/#802/#803/#804). Release will ship as v2.0.0 (v1.70.0 was yanked; its version bump on main is superseded at release time).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012bZreEXUFGmxW1uHBWQMrr

Issue #702, opened 2026-07-27
Membership wave hygiene bundle: cache-key constants, locator stragglers, label dedup, prop hardening


biagiodistefano:
Consolidated non-blocking follow-ups from the follow-up wave's reviews (42fa59c3..cedd1816; all triaged FOLLOW-UP by the wave-final review — details in the SDD ledger):

- [ ] **Adopt the exported cache-key constants at the query definition sites** (`subscription-cache.ts` exports them; `account/memberships/+page.svelte:20,37`, `OrgMembershipInline.svelte:24` still declare inline literals — a future key rename silently no-ops all #693 seeding). Top item.
- [ ] Locator-hoist stragglers: inline `membershipCard` copies in `subscribe-online.spec.ts:133` and `revival.spec.ts:79,123` → `membership-locators.ts`.
- [ ] `getStatusLabel` export in `$lib/utils/subscriptions` — the status→label map now has three copies (StatusBadge, SubscriptionMetrics, SubscriptionsTab filter); exhaustiveness-check `STATUS_ORDER` against the enum while at it.
- [ ] `PaymentsTable`: make `isOnlinePlan` a required prop (optional-defaults-false fails open — a future consumer forgetting it resurrects the 400-producing refund button); gate the online-refund note on a Stripe link actually being present (URL-less dead-end otherwise).
- [ ] `MembershipCard`: `aria-busy` on the billing-portal button (parity with the resume button); #694 failure test should mock the real `ResponseMessage {message}` error shape.
- [ ] Product copy: metrics card `active_count = active + past_due` can sit above a strip reading `Active 10 · Past due 2` with nothing explaining the sum; fr locale mixes "forfait"/"formule" (pre-existing).
- [ ] Manual mobile pass list (from #696's sweep): SubscriptionDrawer, MembershipRequestCard details dialog, ApproveMembershipModal, MakeMemberModal, SubscriptionCreateModal (non-portalled MemberCombobox listbox inside the new scroll container).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #697, opened 2026-07-27
Membership polish batch: auto-eval toast copy, superseded reason key, per-query error gate, E2E locator hoist


biagiodistefano:
Consolidated small follow-ups from PR⑤'s reviews (none release-blocking):

**Product copy**
- [ ] Passing an *automatic* membership questionnaire still shows the generic "Questionnaire submitted — the organization will review it" toast before redirecting (\`/org/[slug]/questionnaire/[id]\` — \`isAutoAccepted\` narrow); for an auto-pass nothing will be "reviewed" — the eligibility gate is already cleared. Tailor the toast (or branch on auto-accept).
- [ ] \`membershipEligibility.reason.membership_questionnaire_pending\` copy is dead on the org-page CTA (the waiting branch renders no note) and is superseded by \`wait.questionnaire_evaluation\` under the resolution order everywhere else — remove or rewire.

**Error handling**
- [ ] \`account/memberships\`: the error branch gates on BOTH queries empty — a partial failure (memberships errored, subscriptions returned a rejoin offer) renders offers with no error line. Per-query gating is the honest shape (PR⑤ Task 2 reviewer).

**Test/E2E hygiene**
- [ ] Hoist the duplicated \`membershipCard\`/\`planCard\`/\`applicationRow\`/\`pageAs\` helpers (5/2/2/4 copies across j23/j27 specs) into \`tests/e2e/support/\`.
- [ ] \`org-policy.spec.ts:103-105\`: one remaining pre-existing comment still claims the false BE-reset mechanism (omitted PUT fields → schema defaults; BE uses exclude_unset) — same one-line reword as the PR⑤ comment wave applied elsewhere.

Context: PR⑤ ledger deferred-minor triage; letsrevel/revel-frontend#692.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

biagiodistefano:
Landed on `feature/membership-subscriptions` in a10b076c (follow-up wave, 42fa59c3..cedd1816 — per-task Opus review + wave-final review clean; gate E2E 491/0, unit 1618/0). Ships with the coordinated FE+BE deploy.

PR #716, opened 2026-07-29
fix(a11y): announce membership load failures, fold review-pending into link name (#687)

### LINKED ISSUE #687: Post-PR④ polish batch: requests/membership a11y, error branches, test hardening, de vocabulary
Consolidated polish batch deferred from PR④'s final whole-branch review (membership-subscriptions launch). All items are small, none release-blocking; they were triaged FOLLOW-UP-ISSUE in the SDD ledger.

**A11y**
- [ ] `MembershipRequestCard`: tier chip has no screen-reader context — add a visually-hidden "Tier:" prefix (member-side precedent: `applications.tierLine`).
- [ ] `MembershipRequestCard`: the "Review pending" hint is a sibling span — attach it to the submission link via `aria-describedby` (or fold into the accessible name) so link-list navigation keeps the state.
- [ ] Spinner idiom: `ApplicationsSection` + `account/memberships` use `role="status"` + `aria-label` on the wrapper; the house's stronger idiom is icon + `<span class="sr-only">` (7 existing sites, e.g. `SeriesPassesTab.svelte:93-96`). Align.

**Error handling**
- [ ] `account/memberships` page has NO error branch — failed memberships/subscriptions queries render the empty state (same bug class fixed for `ApplicationsSection` in PR④).
- [ ] `ApplicationsSection`: a failed background refetch discards already-loaded rows (`isError` replaces the list) — consider `isError && applications.length === 0` gating or an inline banner (invoices-page precedent applies either way).

**Tests**
- [ ] `MembershipRequestsTab`: reject-path toast handling untested (mirror of the tested approve path).
- [ ] `MembershipRequestsTab.test.ts`: page-reset-on-filter-click assertion is vacuous (count:1 fixture → totalPages 1) — needs a count:60 fixture + Next-click arrange to discriminate.

**Locale polish (de)**
- [ ] `applications.loadError` says "Anträge" inside a family that says "Bewerbungen" — align.
- [ ] `membershipRequestsTab.approveFailedGeneric` uses "angenommen" while the approve buttons say "Genehmigen" — align verb pair.

Context: SDD ledger `.superpowers/sdd/2026-07-27-membership-subscriptions-pr4-requests-settings/progress.md` (deferred-minor triage table in the final-review entry).

biagiodistefano:
Closes #687

## Heads-up: most of #687 had already landed

The issue was filed **2026-07-27**, while PR #703 (membership subscriptions, "PR④") was still in review. #703 merged **2026-07-28** — and the fixes were applied *in the branch* before it merged. `git log -S` puts every one of them at commit `20407334` (#703 itself).

Verified item by item against `origin/main`:

| #687 checklist item | State on main | Evidence |
|---|---|---|
| `MembershipRequestCard` tier chip sr-only "Tier:" prefix | ✅ already done | `MembershipRequestCard.svelte`, `membershipRequestCard.tierPrefix` in all 4 catalogs; test *"names the chip for a screen reader instead of reading a bare word"* |
| "Review pending" attached to the submission link | ⚠️ **partly** — see below | `aria-describedby` was wired, but it does not achieve the stated goal |
| Spinner idiom in `ApplicationsSection` + `account/memberships` | ✅ already done | both already `role="status"` + `aria-hidden` icon + `sr-only`, matching `SeriesPassesTab.svelte:93-96` |
| `account/memberships` error branch | ✅ already done, more thoroughly than asked | per-query gating (#697) reports the broken half while rendering the surviving one; 5 tests under *"load failures"* |
| `ApplicationsSection` background-refetch discards rows | ✅ already done | `isError && applications.length === 0`; test *"keeps the loaded rows when a background refetch fails"* |
| `MembershipRequestsTab` reject-path toast test | ✅ already done | two tests under *"reject errors"* |
| Vacuous page-reset assertion | ✅ already done | `count: 60` + `next` + a Next-click arrange |
| de `applications.loadError` "Anträge" → "Bewerbungen" | ✅ already done | reads "Deine **Bewerbungen** konnten nicht geladen werden." |
| de `approveFailedGeneric` "angenommen" → verb-aligned | ✅ already done | reads "konnte nicht **genehmigt** werden", matching "Genehmigen" |

So this PR ships the one item that was **not** actually satisfied, plus the gaps a fresh a11y audit of the same four files turned up.

## What ships

**Review-pending state moved into the link's accessible name** (`MembershipRequestCard.svelte`)

The checklist asked for the hint to attach "via `aria-describedby` (or fold into the accessible name) **so link-list navigation keeps the state**". The shipped `aria-describedby` does not do that: the VoiceOver rotor ("Links"), NVDA's Elements List and the JAWS links list — the way a blind admin actually jumps to a link — expose the accessible **name** and nothing else. An admin landing on the link that way heard only "View questionnaire submission" and could miss that the outcome is provisional.

The caveat now rides in the name via a new localized label. It repeats the visible link text verbatim before adding the caveat, so **WCAG 2.5.3 Label in Name** still holds (voice control's "click View questionnaire submission" keeps working). `aria-describedby` is dropped rather than kept alongside — name, then description, then the visible span's own static text would announce the same three words three times.

**Bare spinner in `MembershipRequestsTab`** — `role="status"`, `aria-hidden` icon, `sr-only` text, matching the house idiom. It was a spinning glyph with no accessible text at all: silent to a screen reader (WCAG 4.1.3). This is the issue's spinner item applied to the sibling file in the same family; the two files it named were already correct.

**`role="alert"` on the error lines** in `MembershipRequestsTab`, `ApplicationsSection` and `account/memberships`. Each replaces the spinner *after* first paint, so without a live region the failure landed silently for anyone who had already moved focus on. The memberships page keeps **two separate alerts** so a partial failure announces only the half that actually broke.

**de noun alignment** — `approve/rejectFailedGeneric` said "Der **Antrag**" while the rest of that admin family says "Anfrage" (`Angefragt`, `Anfragedetails`, "Anfrage von {name} genehmigen"). Same defect class as the issue's locale items, which were themselves already fixed.

## Mutation checks

Every new assertion was deliberately broken, seen to fail, restored, seen to pass — a green test nobody has watched fail is not evidence.

| Mutation | Result |
|---|---|
| strip `role="status"` + `sr-only` from the tab spinner | ✗ *announces that the list is loading* |
| strip `role="alert"` from the tab error | ✗ *announces a failed list load* |
| strip `role="alert"` from `ApplicationsSection` | ✗ *announces the load failure…* |
| strip `role="alert"` from the memberships page (2 sites) | ✗ both new alert tests |
| remove the `aria-label` fold on the card link | ✗ 2 card tests |
| control — all restored | 67/67 pass |

The two **pre-existing** test items from the issue were verified the same way rather than taken on trust:

- removing `requestsPage = 1` from the filter click → *"resets to page 1 when a filter is clicked from a deeper page"* fails with `expected 2 to be 1`. **The de-vacuoused assertion genuinely discriminates.**
- swallowing the reject `onError` toast → **both** reject-path tests fail.

## Verification

- `make check` — 0 errors, and **`✓ Type gate is armed`** (canary passes). Warning count unchanged from the pre-change baseline (377, all pre-existing `state_referenced_locally`).
- `pnpm test` — **173 files, 2106 passed**, 1 todo.
- `svelte-autofixer` — clean on all three components touched.
- E2E not run, per instructions.

## Notes

- Touching `account/memberships/+page.svelte` only in the render path (the error `<p>`s), not the query definitions at the top — that region belongs to the concurrent #702 hygiene bundle. Rebased on `origin/main`; no conflict at time of push.
- The new key `membershipRequestCard.viewSubmissionPendingAria` is in all four catalogs (en/it/de/fr), flat-dotted to match its family.

## Deliberately left undone

The audit surfaced three further items, all real but out of this issue's scope:

1. A single in-flight approve/reject disables the action buttons on **every** rendered card (`isProcessing` is shared), with no announcement — wants per-card scoping.
2. The force-approve dialog disables both buttons *and* suppresses Escape/close while in flight, leaving nothing focusable. Compliant only as "momentary"; a hung request would make it a genuine keyboard trap (WCAG 2.1.2).
3. The force-approve button's spinner has no `sr-only` text. Not added here on purpose: an `sr-only` span inside the button would join its accessible name and break the anchored `name: /^approve anyway$/i` queries — this wants `aria-busy`, not the spinner idiom.

Also observed, not changed: this tab (and `MembersTab`) gate on `isLoading` where `ApplicationsSection` and the memberships page deliberately use `isPending`, to avoid flashing the empty state during auth bootstrap. The admin surface is consistent with itself, so switching one file would be a local deviation — worth a sweep, not a drive-by.


biagiodistefano:
### Review responses

**Copilot — "the visible hint span is announced twice; hide it from AT" (`MembershipRequestCard.svelte:222` and `:406`) — ACCEPTED**, fixed in 6a2f4b0.

Validated before acting rather than taking it on faith. The markup is:

```html
<a aria-label="View questionnaire submission (review pending)">View questionnaire submission</a>
<span>Review pending</span>
```

In browse mode a screen reader announces the link by its accessible name — which now ends in "(review pending)" — and then continues into the sibling span and reads "Review pending" again. So the double announcement is real, and it is exactly the noise I dropped `aria-describedby` to avoid in the first place; leaving the span exposed re-introduced it through the other door. Consistent to fix.

Both spans now carry `aria-hidden="true"`. This is only safe *because* the state lives in the link's accessible name, so the two are pinned together by one test: it asserts the name contains the caveat **and** that every visible hint is `aria-hidden`. Mutation-checked — dropping `aria-hidden` from the spans makes that test fail, restoring it passes.

Note the existing `getAllByText('Review pending')` assertion still holds: `getByText` walks the DOM, not the accessibility tree, so it keeps guarding the hint's visual presence for sighted users.

**CodeRabbit — rate-limited**, no review produced ("Review limit reached… next review available in 52 minutes"). Not blocking on it.

Re-verified after the fix and after rebasing onto `9c861e50` (#702, which moved into the same two files): `make check` 0 errors with `✓ Type gate is armed`, `pnpm test` 173 files / 2108 passed.

PR #718, opened 2026-07-30
refactor(events): rewire address_visibility + pronoun distribution into visibility_settings (#713)

### LINKED ISSUE #713: refactor(events): rewire address_visibility + pronoun distribution into visibility_settings (phase 2)
Frontend counterpart of letsrevel/revel-backend#793 (visibility settings phase 2). Follow-up to #690, which shipped the phase-1 toggles and explicitly deferred this rewiring.

**Breaking change — ships in a single coordinated release with the backend.** There is no deprecation window: the backend drops both columns and both top-level schema fields in one PR, so this must merge and deploy together.

## What changes

Two fields move out of the top level of the event schemas and into the existing `visibility_settings` object:

| Removed from the wire | New location |
|---|---|
| `address_visibility` | `visibility_settings.address_visibility` |
| `public_pronoun_distribution` | `visibility_settings.show_pronoun_distribution` (**renamed**) |

Values, enum members (`public` / `unlisted` / `private` / `members-only` / `staff-only` / `attendees-only`) and defaults (`public`, `false`) are unchanged — only the nesting and the pronoun key's name.

`visibility_settings` is already round-tripped by every form touched here (#690), and it is **merged, not replaced**, server-side: sending `{"address_visibility": "staff-only"}` leaves the other toggles alone. So these two just become two more keys in an object the forms already send.

## Prerequisite

- [ ] Backend PR merged, then `pnpm generate:api`. Every type error the regen produces is a site on this list.

## Call sites (40 references across 10 files)

**Organizer forms — the bulk of the work:**

- [ ] `src/lib/components/events/admin/event-payload.ts` — 4 serialization sites (L101, L141, L176, L218) plus the `public_pronoun_distribution` write at L229 and the two form-type declarations at L18/L22. All five `?? 'public'` / `|| 'public'` fallbacks move inside the nested object.
- [ ] `src/lib/components/events/admin/event-payload.test.ts` — L35, L61 fixtures.
- [ ] `src/lib/components/events/admin/EventEditor.svelte` — form-state seeding at L118/L134, type declarations at L95/L484.
- [ ] `src/lib/components/events/admin/EventWizard.svelte` — L134 seeding, type declarations at L118/L451.
- [ ] `src/lib/components/events/admin/LocationSection.svelte` — the address-visibility `<select>` itself (L268-270) plus defaults at L88/L107 and the type at L24.
- [ ] `src/lib/components/events/admin/DetailsStep.svelte` — the pronoun-distribution checkbox (L566-567) and the types at L45/L73.
- [ ] `src/lib/components/event-series/admin/TemplateEditDialog.svelte` — the densest one: seeding (L83, L136, L146) **and** the dirty-diff logic (L182-183, L201-202). The diff compares against `original.address_visibility ?? 'public'`; both sides move into the nested object.
- [ ] `src/lib/components/event-series/admin/template-edit-fields.ts` — L27 union member and the L84 field descriptor.
- [ ] `src/lib/components/event-series/admin/RecurringEventWizard.svelte` — L51 type, L75 default.

**Public event page:**

- [ ] `src/routes/(public)/events/[org_slug]/[event_slug]/+page.svelte` — `showPronounDistribution={event.public_pronoun_distribution || data.isOwner || data.isStaff}` at **L509 and L579** (mobile + desktop mounts, identical props — change both). Becomes `event.visibility_settings.show_pronoun_distribution || …`.

## Behavior change: pronoun distribution no longer 403s

`GET /events/{id}/pronoun-distribution` converges onto the same redaction pattern as the phase-1 count gates. A caller who may not see the distribution now gets **`200` with an empty body** instead of `403`:

```json
{ "distribution": [], "total_with_pronouns": null, "total_without_pronouns": null, "total_attendees": null }
```

Low impact in practice — the page already pre-gates client-side with `showPronounDistribution`, so the 403 was rarely reached — but worth confirming:

- [ ] Nothing in the fetch path treats a `403` from this endpoint as a meaningful signal.
- [ ] `PronounDistributionChart.svelte` renders acceptably for an empty `distribution` with null totals (it takes `totalWithoutPronouns` and interpolates `pronounDistribution.summary`), or the caller skips mounting it when `total_attendees` is null.

## `TemplateEditDialog` will 422, not fail silently

`TemplateEditSchema` is `extra="forbid"` on the backend. A stale build that PATCHes `address_visibility` at the top level gets a **422**, unlike the event-edit endpoint where an unknown field is silently dropped. That is the intended loud failure — it just means the template dialog is the first thing to check if the release skews.

## Testing

- [ ] `pnpm check` clean after regen (the type errors are the checklist).
- [ ] `pnpm test` — `event-payload.test.ts` updated.
- [ ] Manual: create + edit an event with a non-public address visibility; confirm it persists and that changing it does not reset the phase-1 toggles (`show_capacity` etc.).
- [ ] Manual: template edit dialog round-trips both fields and the dirty-diff still suppresses no-op saves.
- [ ] Manual: pronoun distribution section on the public event page, opted in and opted out, as anonymous / attendee / organizer.

## Out of scope

- Backend work — tracked in letsrevel/revel-backend#793.
- Any change to what the toggles *mean*. This is a pure relocation plus one rename.

biagiodistefano:
Frontend counterpart of letsrevel/revel-backend#793 (visibility settings phase 2).

**Breaking, coordinated change — must deploy together with the backend PR.** There is no deprecation window: the backend drops both columns and both top-level schema fields in one PR.

## What changes

| Removed from the wire | New location |
|---|---|
| `address_visibility` | `visibility_settings.address_visibility` |
| `public_pronoun_distribution` | `visibility_settings.show_pronoun_distribution` (**renamed**) |

Values, enum members (`public` / `unlisted` / `private` / `members-only` / `staff-only` / `attendees-only`) and defaults (`public`, `false`) are unchanged — only the nesting and the pronoun key's name. `visibility_settings` is merged server-side at sub-key granularity, so these two just become two more keys in an object every touched form already round-trips.

## How it's structured

Seven commits, each independently reviewable:

1. `53b6c01f` — regenerate the API client (the 18 resulting type errors were the checklist)
2. `f7cbccf7` — `event-visibility.ts` models all five keys
3. `7430ff50` — payload builders stop sending the flat fields
4. `fc55fea2` — event admin forms (`EventEditor`, `EventWizard`, `LocationSection`, `DetailsStep`)
5. `7741e5ad` — event-series admin (`TemplateEditDialog`, `template-edit-fields`, `RecurringEventWizard`)
6. `5418c81d` — public event page pronoun gate
7. `4035bec3` — review polish
8. `3b9cf07e` — review response: type tightening + `TemplateEditDialog` component test
9. `69571f1c` — E2E coverage

### The type model

`src/lib/utils/event-visibility.ts` now carries a deliberate three-way split, because the five keys are not interchangeable:

- **`ResolvedVisibilityToggles`** — the three preset toggles only. This is what `EventVisibilityFields` emits, and what `VISIBILITY_PRESETS` describes. `address_visibility` is an enum edited in `LocationSection` and the pronoun flag is a checkbox in `DetailsStep`; neither belongs to a preset, so `matchVisibilityPreset` / `isNonDefaultVisibility` still iterate only these three (a non-default address visibility must not make the preset row read "custom", nor auto-open the visibility accordion).
- **`ResolvedVisibilitySettings`** — all five resolved keys. `resolveVisibilitySettings` and `diffVisibilitySettings` cover all of them, or the new keys would be dropped on write and invisible to the template dialog's dirty-diff.
- **`VISIBILITY_PRIVILEGED`** — what an owner/staff viewer sees, deliberately *not* `VISIBILITY_DEFAULTS`. The pronoun key's default is opt-in `false`, but a privileged viewer always sees the distribution — which is exactly what the old `event.public_pronoun_distribution || isOwner || isStaff` encoded.

### Merge, never replace

Because the preset UI emits only a *partial* of `visibility_settings`, every write site had to change from assignment to a merge. `DetailsStep.svelte`'s `onChange` and `TemplateEditDialog.svelte`'s were both live bugs waiting to happen the moment the two relocated keys shared the object: they would have silently wiped the organizer's address visibility and pronoun opt-in. Both now spread. All six write sites across `src/` were swept and verified.

### One behavior change worth calling out

This is otherwise a pure relocation, with one exception. Before this PR, only `buildEditorUpdateData` sent `public_pronoun_distribution` — `buildWizardStep1UpdateData`, `buildWizardStep2UpdateData` and `buildRecurringTemplateCreateData` all dropped it, so **the pronoun-distribution checkbox was inert in the create wizard and the recurring-template wizard**. Now that the value rides inside `visibility_settings`, all four builders carry it and those two flows honour the checkbox for the first time. That is a latent-bug fix rather than a regression, but it is a real behavior delta and shouldn't surprise anyone in review.

### Pronoun distribution no longer 403s

`GET /events/{id}/pronoun-distribution` converged onto the same redaction pattern as the phase-1 count gates: a caller who may not see the distribution now gets `200` with an empty body and null totals instead of `403`. Audited, no frontend change needed — nothing in the fetch path treated `403` as a signal (the generated SDK declares a 200-only response), and `AttendeeList.svelte` already null-guards all three totals and falls through to the "no attendees" copy without ever mounting `PronounDistributionChart`, so an empty response cannot produce a zeroed or NaN-bearing panel.

## Testing

- `make check` — **0 errors** (from 18 after regen), 374 warnings (all pre-existing `state_referenced_locally`; 3 fewer than `main`). Type-gate canary confirms the gate is armed. i18n, file-length, no-SSR-token-leak and image-alt audits clean.
- `make test` — **173 files, 2120 passed**, 1 todo, 0 failed.
- `event-payload.test.ts` now asserts the *negative* across all four builders (`not.toHaveProperty('address_visibility')`, same for the pronoun key). That absence is the load-bearing assertion: `TemplateEditSchema` is `extra="forbid"` and 422s on a stray top-level key, where the event-edit endpoint would silently drop it.
- `event-visibility.test.ts` covers the widened defaults, privileged-viewer resolution, five-key diffing, and preset-matching's indifference to the two non-preset keys.

### Verified live against the running backend

Beyond the automated gates, I smoke-tested the wire contract against a locally-running backend on the matching `refactor/793-visibility-settings-phase2` branch:

- The backfill migration preserved existing values (seeded events came back with `address_visibility` of `private` / `staff-only` / `unlisted`), and no event schema still carries either field at the top level.
- **Sub-key merge holds:** `PUT {"visibility_settings": {"address_visibility": "members-only"}}` leaves a previously-set `show_capacity: false` untouched, and vice versa. This is the property the whole PR rests on.
- Legacy top-level keys are silently dropped by the event-edit endpoint (as documented) rather than honoured — so a stale client degrades to "no change", not to a wrong value.
- The pronoun endpoint's new redaction behaviour, confirmed per viewer class:

  | viewer | `show_pronoun_distribution` | response |
  |---|---|---|
  | plain user | `true` | `200`, real numbers |
  | plain user | `false` | `200`, `distribution: []`, all three totals `null` |
  | owner | `false` | `200`, real numbers (privileged bypass) |
  | anonymous | either | `401` (endpoint requires auth) |

### E2E coverage

This surface had **no E2E coverage at all** — phase 1 (#690) left the gap, so these specs cover the phase-1 toggles too, not just the two relocated fields. 6 tests, green on both `chromium` and `mobile-chrome`, 0 skipped:

- `j10-event-admin/visibility-settings.spec.ts` — a three-round admin round-trip proving **mutual non-clobbering**: set a non-public address visibility and opt into the pronoun distribution, then flip an unrelated phase-1 toggle and confirm both survive, then change the address visibility again and confirm the phase-1 toggle survives that. This is the assertion that would have caught the merge-vs-replace bug.
- `j05-rsvp/visibility-disclosure.spec.ts` — public-page disclosure: the pronoun section appears for a non-privileged viewer only when opted in, and appears for the **owner even when opted out** (the `VISIBILITY_PRIVILEGED` bypass); plus `show_attendee_count: false` hiding the count and `show_attendee_list: false` rendering the hidden-list copy instead of a real attendee's name.

Both specs were proven falsifiable: reverting the `DetailsStep` merge to a bare assignment makes the admin spec fail, and it passes again on revert. Likewise the `TemplateEditDialog` component test's preset-merge case fails on that regression and only on that one.

The E2E suite still isn't wired into CI (#593), so these run locally.

Closes #713


biagiodistefano:
Thanks both — addressed in `3b9cf07e`.

**@coderabbitai — taken.** `TemplateEditDialog` had no component test at all, and it is the densest site on this branch (it seeds both relocated fields, drives the dirty-diff, and emits a 422-sensitive patch). Added `TemplateEditDialog.test.ts` with four cases: seeding from nested `visibility_settings`, the emitted patch nesting both values *and* carrying no top-level `address_visibility`/`public_pronoun_distribution` (the `extra="forbid"` guard), preset changes preserving both relocated values, and no-op suppression. The merge-preservation case was verified by temporarily regressing the `onChange` merge back to an assignment and confirming that test — and only that test — fails.

**@copilot-pull-request-reviewer — not taken; the premise is incorrect.** Object spread does not throw on `null`: it invokes CopyDataProperties, which ignores both `null` and `undefined` (only *array* spread/iterable destructuring throws). Verified directly:

```
$ node -e 'console.log(JSON.stringify({...null, a:1}))'
{"a":1}
```

So none of the flagged spreads can throw, and adding `?? {}` fallbacks would be dead code. `DetailsStep.svelte` additionally never declares `| null` on that field, so that half of the finding has no basis.

What *was* real is the type imprecision that provoked it, which a human reviewer independently flagged: `LocationSection.svelte` declared `visibility_settings?: EventVisibilitySettings | null` although no event schema is nullable — `EventCreateSchema`, `EventEditSchema` and `EventDetailSchema` are all `?: EventVisibilitySettings`, and only `TemplateEditSchema` is `| null` (it does not flow into this component). Dropped the `| null` so the prop type matches what callers actually pass, which removes the false-positive surface rather than papering over it.

Issue #505, opened 2026-06-28
i18n: landing page flashes to English on first visit and "Welcome to" heading stays stuck in Accept-Language locale


biagiodistefano:
## Summary

On a **fresh visit** (best reproduced in a private/incognito window) with a browser whose `Accept-Language` is **not** English (e.g. French), the landing page is server-rendered in that language, then the whole UI **flips to English** on hydration — **except the "Welcome to Revel" `<h1>` heading**, which stays stuck in the `Accept-Language` language (e.g. "Bienvenue sur Revel") on every refresh.

This is actually **two bugs stacked on top of each other**:

1. A **locale propagation bug** that makes the entire UI flash from the `Accept-Language` locale to English on first load.
2. A **`{@html}` hydration bug** that makes the welcome heading (and only the heading) stay frozen in the stale server-rendered locale.

## Reproduction

1. Open a **private/incognito** window (so no cookies exist).
2. Set the browser's preferred language to **French** (or German/Italian).
3. Visit the landing page `/`.
4. Observe: the page briefly renders fully in French (SSR), then everything switches to English on hydration **except** the `<h1>` which remains **"Bienvenue sur Revel"**.
5. Refresh — the mismatch reproduces every time.

Using the in-app language switcher **masks** the bug (see "Why it's hard to notice" below), which is why it only shows up on first/private visits.

### Bug 1 — Server-detected locale is never propagated to the client Paraglide runtime

`src/lib/i18n.ts` (`i18nHandle`) detects the locale server-side from URL → cookie → `Accept-Language` → default, then:
- calls `setLocale(lang)` for the SSR render, and
- writes a **`user_language`** cookie (`src/lib/i18n.ts:61`), and
- injects `%paraglide.lang%` into `<html lang>`.

But the generated Paraglide client runtime resolves the locale with the strategy `["cookie", "globalVariable", "baseLocale"]` (`src/lib/paraglide/runtime.js:35-39`) where the **cookie strategy reads `PARAGLIDE_LOCALE`** (`src/lib/paraglide/runtime.js:25`), **not** `user_language`.

So on a fresh visit:
- **SSR**: `getLocale()` → `fr` (via the server `setLocale`) → page rendered in French.
- **Client hydration**: `PARAGLIDE_LOCALE` cookie is absent → `globalVariable` (`_locale`) is unset → falls back to **`baseLocale` = `en`**.

[...]

PR #764, opened 2026-08-03
feat(landing): poster redesign — seven-panel color-block homepage (#763)

biagiodistefano:
* feat(landing): add poster panel color tokens with AA audit coverage

* feat(landing): add poster landing page copy for all six locales

Adds the `home.poster.*` block (64 keys) to en, it, de, fr, es and pt catalogs for the poster-style landing page redesign.

* fix(landing): correct inverted pricing claim in it poster copy

`home.poster.pricingCompare` used "costerebbe", making the €15 ticket the subject of the cost verb and reading as if the ticket itself would cost €1.79 on Eventbrite. Reworded so Eventbrite is the explicit subject of the charging verb, preserving the fee-on-ticket meaning like every other locale.

* feat(landing): add PosterPanel and PosterSticker primitives

PosterPanel wraps a full-bleed poster section and optionally renders a diagonal cut strip that reveals the next panel's color. PosterSticker is the white (or ink-inverted) rounded label, with rotation clamped to the [-3, 3] degree tilt discipline.

The .cut-* helper classes live in app.css as plain global CSS, outside @layer base: they are applied through a dynamic prop, so Tailwind's content scan would otherwise purge them.

* fix(landing): make PosterPanel cutToClass a compile-checked union

cutToClass was typed `string` and documented as "Background class of the NEXT panel", which invites a bg-[...] value. That value type-checks and renders, but leaves --cut-color unset, so the gradient resolves to an invalid color and the diagonal seam silently disappears.

Close the type to the six .cut-* helpers that actually exist in app.css, so a wrong value is a build error instead of an invisible rendering bug, and correct both the JSDoc and the <style> comment to describe the real contract. Verified with a throwaway canary component: passing a bg-[...] class is now rejected by svelte-check.

Seven downstream tasks consume this prop.

* feat(landing): add RotatingNoun flip device (SSR-safe, reduced-motion aware)

The "living headline" device for the poster landing page: stacks all items in an inline-grid cell so the box is as wide as the longest item and swapping never reflows the headline. Index 0 is what SSR renders and rotation starts only in onMount, so hydration never sees changed markup. prefers-reduced-motion keeps item 0 static. The element is aria-hidden; callers own the accessible sentence.

* fix(landing): stub matchMedia globally and key RotatingNoun by index

Two review findings on the RotatingNoun device:

jsdom implements no matchMedia, and RotatingNoun queries it for prefers-reduced-motion on mount — so any future test rendering it or a parent of it (Task 5's HeroPanel) threw "window.matchMedia is not a function" unless it knew to stub it. Add a stub to vitest.setup.ts beside the ResizeObserver polyfill, defaulting to matches:false, install-only- when-absent like its neighbours. Per-test vi.stubGlobal still overrides it, which is what keeps the reduced-motion branch deterministic.

Key the #each by index rather than by the item string: a locale whose noun list repeats a word (or a degenerate endings list like ['a','a']) threw each_key_duplicate at runtime, breaking the landing page for that locale alone. The list is static and index-stable, so an index key is equivalent to unkeyed while keeping svelte/require-each-key satisfied.

Full unit suite green: 188 files, 2369 passed.

* feat(landing): hero panel with rotating noun and inclusive kicker

HeroPanel composes PosterPanel + PosterSticker + RotatingNoun into the poster hero: a gradient crimson→purple panel, the tilted RevelMark sticker, the rotating-noun headline, and auth-aware CTAs.

The full headline is exposed to assistive tech as a single static sr-only string (`home.poster.heroAria`); the animated variant is aria-hidden, so screen readers never hear a word swap mid-sentence.

The inclusive rotating-vowel kicker for it/es/pt is migrated from the retired LandingHero (#505), with the inclusive marker FIRST so SSR and no-JS render the inclusive form rather than a gendered one.

Contrast: the subline and the "already have an account" link use full-opacity white. At their sizes (text-lg medium, text-sm) a 0.85–0.92 alpha over the gradient drops below the 4.5:1 AA threshold for non-large text.

Also fixes the RotatingNoun flip rendering flat: `perspective` on an ancestor only reaches its direct children, and the `.rotating-noun` wrapper in between defaults to `transform-style: flat`, which flattens the flipping spans back out. The bridging rule lives in app.css so every panel embedding the device gets the same 3D behaviour.

* fix(landing): AA-safe ghost CTAs, truthful kicker comment, tighter hero tests

Review follow-up on the poster hero, three issues.

The ghost CTAs failed WCAG AA. White bold 16px over a white/0.16 wash measures 4.14:1 on the purple end of the gradient and 3.86:1 on the crimson end, short of the 4.5:1 that size requires. Tinting the fill with ink instead of white takes it to 7.19:1 / 6.32:1, keeping the white border and label. Both call sites carry the numbers inline, because the effective color is composited from a translucent fill over a gradient at paint time and so is invisible to audit-brand-themes.py.

The INCLUSIVE_KICKER comment claimed the block was migrated verbatim from the retired LandingHero, which was wrong in a way that mattered: the #505 `browser` gate was deliberately dropped so SSR renders the inclusive-first ending rather than nothing, and the connector words went with it. The comment now says so, and points at the real hydration safeguard — the Paraglide locale cookie written in src/lib/i18n.ts.

Tests under-asserted. The logged-in case now pins the auth switch as exclusive, and two new cases pin the link contracts: the log-in prompt is a real link to /login, and the demo CTA opens in a new tab with rel=noopener rather than handing a cross-origin page window.opener. Each new assertion was mutation-checked — breaking the corresponding source turned that test, and only that test, red.

* feat(landing): amber pricing panel with fee calculator hook

Adds PricingPanel — the €0 headline slab on amber, the "no hidden fees" sticker, the Eventbrite comparison line and the button that hands the FeeCalculatorModal back to the page (the page keeps owning that state).

All copy runs through existing message keys, so the panel is translated in all six locales on arrival.

Contrast note: the brief dimmed the label with `opacity-65` and the two footnotes with `opacity-75`. Both are dropped in favour of full-opacity ink. Ink on amber is 9.42:1 solid; at 65% it composites to roughly 3.4:1, which fails AA for the text-sm label. Composited alpha is also invisible to scripts/audit-brand-themes.py, so the failure would never surface in the gate. Hierarchy is carried by weight, size and tracking instead.

* feat(landing): communities panel with questionnaire mock

CommunitiesPanel (periwinkle, cuts to crimson) pairs the "too kinky for Eventbrite" headline and vetting/privacy pills with QuestionnaireMock, a decorative card exposed as a single role="img" so screen readers get one accessible name instead of walking the fake form.

AA: the small uppercase label carries no opacity — ink at 65% composites to ~2.5:1 on periwinkle, a failure at text-sm. Pills tint only the background (ink@12%), so their solid-ink text stays ~8:1. The mock's ink@72% option text sits on the white card (~9.7:1), hand-verified since composited alpha is invisible to audit-brand-themes.py.

* feat(landing): venues panel with seat map mock

Crimson-deep panel for venues/theaters with a decorative seat-map card (single role="img", aria-hidden internals, deterministic taken seats so SSR and client agree).

AA note: white on crimson-deep measures 4.58:1 — over the floor with almost no headroom — so no white string on this panel is alpha-softened (the label drops the briefed opacity-75, which would composite to 3.08:1). Composited alphas that scripts/audit-brand-themes.py cannot see are hand-verified inline.

* feat(landing): features panel with ticket stubs and SEO link strip

Four dashed ticket-stub cards on the lavender paper panel plus the six locale-prefixed SEO landing links, preserving the link contract of the retired LandingUseCases section.

AA notes: the small uppercase labels stay full-opacity ink (softened 12px text on light panels is where this poster keeps failing AA); the stub body's ink@72% sits on WHITE cards and measures 6.8:1, verified by hand. Tailwind's default spacing scale has no 4.5 step, so the stub padding is p-[18px].

* feat(landing): open source and close panels

OpenSourcePanel (ink) and ClosePanel (purple→crimson-deep gradient) close the poster stack. ClosePanel swaps its primary CTA to Browse Events when org creation is gated off.

Contrast is documented inline with measured ratios, since alpha-composited colors are invisible to audit-brand-themes.py: white/0.5 label on ink 5.19:1, white/0.8 body 11.43:1, white on the close gradient 5.52:1 (purple end) to 5.38:1 (far end, #C42D61) — so the ghost CTA needs no ink wash here, unlike the hero's. The close tagline stays full-opacity white: at 18px medium a 0.9 alpha leaves only 4.62:1 at the far end.

* test(landing): assert the close-panel gate in both directions

The canCreateOrg=true case never proved "Browse Events" was absent, so a ClosePanel that let the fallback link escape the {:else} — gating only create-org — passed the whole suite. Mutation-checked: with that leak applied, this assertion is the only one that goes red.

* feat(landing): compose poster homepage, retire card-grid landing

Replace the card-grid homepage with the seven poster panels in cut order (Hero -> Pricing -> Communities -> Venues -> Features -> OpenSource -> Close) and delete LandingHero/LandingUseCases, which nothing else imported.

Visual-smoke fixes found while composing:
- The hero R sticker anchored to PosterPanel's full-bleed <section>, so past ~1440px it pinned to the viewport edge and clipped its heart. It now rides a relative wrapper inside the max-w-6xl content box.
- min-w-[300px] on the Communities/Venues/OpenSource text columns forced a 300px track inside a 288px container at 320px, eating the panel padding. Now min-w-0 with the 300px floor from sm up; OpenSource also takes basis-full below sm so the R mark wraps onto its own row instead of squeezing the column to 143px.

* fix(landing): keep the poster text columns full-width below sm

`min-w-[300px]` was doing double duty on the Communities/Venues text columns: a width floor AND the thing that forced the flex wrap on mobile. Swapping it for `min-w-0 sm:min-w-[300px]` (to stop the 300px track overflowing a 288px container at 320px) removed the wrap trigger, so in the 390-639px band the mock still fit on the first line and `flex-1` handed the text column ~2px — invisible because PosterPanel's `overflow-hidden` clipped the mangled text instead of showing a scrollbar.

Adds the `basis-full sm:basis-0` companion that OpenSourcePanel already carried. Measured text-column widths: 358px @390, 382px @414, 398px @430, 607px @639; 288px @320 with symmetric padding; desktop unchanged at 1280 (side-by-side, 732px / 692px).

* docs(changelog): record the poster landing-page redesign

Adds the [Unreleased] > Changed entry for the full-bleed color-block poster homepage that replaces the card-grid landing page.

Quality gates verified green on this branch: make fix (no-op), make check (exit 0, svelte-check 0 errors, type canary armed), audit-brand-themes.py (0 WCAG failures, 0 confusables), vitest (2385 passed).

The e2e copy-assertion sweep is intentionally excluded from this commit and held for a separate session; a report-only grep found zero specs asserting retired homepage copy.

* chore(i18n): drop message keys orphaned by the poster landing page

The card-grid homepage deleted in 25305dab was the only consumer of 27 keys in the `home` namespace (welcome/tagline/keyword, the features and use-case blocks, and the try-the-demo copy). Remove them from all six catalogs.

Verified orphaned before removal: no `home.<key>` reference and no bare name reference anywhere in src/, no dynamic `m[` template-literal builds a home key, and the `case 'home'` branch in src/lib/seo/build.ts uses hardcoded literals rather than messages. `home.alreadyHaveAccount` and the whole `home.poster` subtree stay — they back the new landing page.

Survivor order is untouched (fr.json included); the diff is deletions only. paraglide:compile, i18n-check (7976 keys, 100% parity), vitest (2385 passed) and svelte-check (0 errors) are green.

* fix(i18n): restore the EEA/UK card-fee split in the fee disclosure

Retiring the old pricing cards for the poster landing page dropped the UK card rate from the product entirely: the surviving FeeCalculatorModal string quoted only the EEA rate, so a UK organizer was shown the wrong number.

Amend the existing learnMore.feeCalculator.creditCardFeesDescription key in all six catalogs rather than reintroducing a key, so each locale keeps its own register and trailing "these go to Stripe" clause. Euro figures stay verbatim in each catalog's existing convention, and the European Economic Area uses each locale's official name (SEE / EWR / EEE) rather than a calque.

FeeCalculatorModal's own source comment already recorded that UK cards cost more; only the user-facing copy had lost it.

* refactor(landing): scope preserve-3d, index-key the stubs, cover OpenSourcePanel

Three review follow-ups on the poster panels:

- Move .rotating-noun's transform-style:preserve-3d out of app.css into a scoped style block in RotatingNoun.svelte, carrying the why-comment across. The class is a literal in that component's markup, so Svelte scoping and Tailwind purging both see it — unlike the .cut-* helpers, which are composed dynamically and stay global. The perspective still lives on HeroPanel's ancestor and preserve-3d lands on the same element, so the flip keeps its depth (verified in-browser: computed preserve-3d, parent perspective 800px, and the item box widening 220.7 -> 225.2px at rotateX(60deg) where a flat wrapper moves it the other way).

- Key FeaturesPanel's stub loop by index instead of by the translated title, so a future locale rendering two titles identically cannot throw each_key_duplicate and take down the homepage. Same reasoning, and same comment style, as RotatingNoun's item loop.

- Add OpenSourcePanel.test.ts covering the h2, the GitHub CTA's href/target/ rel=noopener, and that the decorative star stays out of the accessible name. That last assertion was confirmed failing-first by temporarily stripping the aria-hidden, so it pins the behaviour rather than passing vacuously.

* feat(landing): styleguide wordmark, gating questionnaire, pricing and copy polish

- ClosePanel lockup per the Digital Brand Styleguide: let's Light, revel Semibold (amber period included), tracking 60 on the wordmark only
- Styleguide rules recorded permanently in CLAUDE.md (wordmark, gradient, safe zones, BBH Bartle social-only, picture do/don'ts); local PDF gitignored
- Copy trimmed across all six locales: hero subline and panel bodies shortened, Eventbrite comparison dropped, close tagline says high fees
- Pricing: fee figure semibold via question/rest i18n split; new self-managed payments line (cash, wire transfer, free dashboard)
- Self-hosting docs link moved to the open source panel (See how)
- Questionnaire mock: Approve/Reject buttons matching the product, vouch question for a more gating door-policy feel
- Venues and communities panels swapped; cut chain rewired to match

---------

## Summary by CodeRabbit

* **New Features**
  * Redesigned the landing page with a poster-style layout, rotating headlines, pricing details, community and venue sections, feature highlights, open-source information, and clear calls to action.
  * Added localized landing-page content and visual previews for venue seating and community questionnaires.
  * Updated fee information to distinguish EEA and UK card rates.

* **Accessibility**
  * Improved contrast, accessible labels, secure external links, and reduced-motion behavior.

* **Documentation**
  * Updated brand typography and digital styleguide guidance.


biagiodistefano:
### Code review

Found 3 issues:

1. Pull-up opacity rule violated on `create-org` — the `-mt-12` pull-up over the `bg-secondary` band lands on a translucent `bg-highlight/20` block in the `ownsOrganization` branch (CLAUDE.md says "**Pull-up opacity rule:** whenever body content is pulled up to overlap a band's bottom edge, the first block it lands on must be an opaque surface (never `bg-card/NN`)"). The same PR fixes this exact defect elsewhere (`PollPrivacySummary` `bg-card/50`→`bg-card`, opaque wrappers added to the dashboard cards).

https://github.com/letsrevel/revel-frontend/blob/709bc360a3202d2ed794cbeabf0c7ed9803263d5/src/routes/%28auth%29/create-org/%2Bpage.svelte#L122-L129

2. Same rule, second branch — the `!user?.email_verified` card is `bg-destructive/10`, also translucent, also the first block on the pull-up.

https://github.com/letsrevel/revel-frontend/blob/709bc360a3202d2ed794cbeabf0c7ed9803263d5/src/routes/%28auth%29/create-org/%2Bpage.svelte#L156-L160

3. Three public pages document the `secondary/55` (dark `/28`) content wash as "a composited alpha is invisible to `scripts/audit-brand-themes.py`" and quote hand-computed ratios — but that pair **is** registered in `COMPOSITED_PAIRS` ([audit-brand-themes.py#L260-L266](https://github.com/letsrevel/revel-frontend/blob/709bc360a3202d2ed794cbeabf0c7ed9803263d5/scripts/audit-brand-themes.py#L259-L267)) and the script prints `12.36 / 6.43 / 4.97` (light) and `15.68 / 7.44 / 6.30` (dark), not the `12.42 / 6.45` and `15.75 / 7.47` in these comments. CLAUDE.md says "Let the script print the ratio and paste **that** into the code comment, never the reverse: hand-written ratios were wrong repeatedly during the 2026-08 rebrand (#783)". `LandingPageTemplate.svelte` in this same PR already carries the correct script-sourced numbers for the identical composite, so the three below just need to match it.

https://github.com/letsrevel/revel-frontend/blob/709bc360a3202d2ed794cbeabf0c7ed9803263d5/src/routes/%28public%29/events/%5Borg_slug%5D/%5Bevent_slug%5D/%2Bpage.svelte#L346-L358

Also at [`org/[slug]/+page.svelte#L240-L248`](https://github.com/letsrevel/revel-frontend/blob/709bc360a3202d2ed794cbeabf0c7ed9803263d5/src/routes/%28public%29/org/%5Bslug%5D/%2Bpage.svelte#L240-L248) and [`series/[series_slug]/+page.svelte#L153-L162`](https://github.com/letsrevel/revel-frontend/blob/709bc360a3202d2ed794cbeabf0c7ed9803263d5/src/routes/%28public%29/events/%5Borg_slug%5D/series/%5Bseries_slug%5D/%2Bpage.svelte#L153-L162).

biagiodistefano:
Follow-up for the review findings above: #800 (this PR had already merged when the review landed). Covers all three reported issues plus the toolbar raw-hue sweep and the EmptyState ratio comments.

PR #820, opened 2026-08-07
fix(csp): stop mode-watcher emitting a nonce-less inline script

### LINKED ISSUE #816: csp: mode-watcher emits a nonce-less inline script — violation logged on every page
Found during the #812 pre-merge smoke pass (visible on 100% of page loads, prod build).

**Symptom:** every route logs `Executing inline script violates ... 'script-src 'self' 'nonce-…'`. The served HTML contains the two nonce-carrying scripts from `app.html` plus one **bare** `<script>` with no nonce — the `setInitialMode(...)` block emitted by `<ModeWatcher />` (`src/routes/+layout.svelte`). Its FOUC-prevention never executes; the app's own nonced theme script in `app.html` appears to cover the same job (dark-mode FOUC regression specs pass), so visible impact is nil — but it's a guaranteed console error and CSP-report noise in prod.

**Fix sketch:** disable mode-watcher's injected script (`nonce`/`disableTransitions`-style options — it accepts a `nonce` prop; pass `%sveltekit.nonce%`-derived nonce from the layout data if available) or drop the component's script injection entirely and rely on the app.html script, documenting that dependency. Cf. the `fe-csp-inline-script-nonce` rule: inline scripts live in `app.html` with the nonce.

biagiodistefano:
## Symptom

Every route in the production build logged, once per page load:

```
Executing inline script violates the following Content Security Policy directive
'script-src 'self' 'nonce-idTh8tvLr2IxELsfy3eOQA=='. Either the 'unsafe-inline' …
```

The served HTML carried the two nonce-carrying scripts from `app.html` plus one **bare** `<script>` with no nonce.

## Root cause

`<ModeWatcher />` (mode-watcher 1.1.0) renders its anti-FOUC snippet through `{@html}` in `<svelte:head>`:

```svelte
{@html `<script${trueNonce ? ` nonce=${trueNonce}` : ""}>(` + setInitialMode.toString() + …}
```

With no `nonce` prop set, that is a bare inline `<script>` — which our `script-src 'self' 'nonce-…'` policy refuses. Confirmed by grepping the served HTML:

```
$ curl -s http://localhost:4173/ | grep -o '<script>[^<]\{0,160\}'
<script>(function setInitialMode({ defaultMode = "system", themeColors, darkClassNames = ["dark"], …
```

Its `nonce` prop is **not** a usable escape hatch here: SvelteKit's per-response nonce is only interpolated into `app.html` (`%sveltekit.nonce%`) and is never exposed to components or `load` functions, so there is nothing to thread through.

Visible impact was nil, because the snippet never executed and the nonced anti-FOUC script already in `app.html` (#440) does the identical job — same `mode-watcher-mode` storage key, same `dark` class, same `colorScheme`. So this was pure console-error and CSP-report noise.

## Fix

`<ModeWatcher disableHeadScriptInjection />`. mode-watcher documents that prop for exactly this case ("Set this if you are manually injecting the script using a hook"), and it swaps `ModeWatcherFull` for `ModeWatcherLite`, which emits no script at all. Both `app.html` and `+layout.svelte` now carry a comment recording that `app.html` is the **only** anti-FOUC script and the two must stay in sync.

Zero new dependencies; three files touched (layout, app.html, one regression spec).

## Evidence

Production build (`make build`) served by adapter-node, before vs. after:

```
--- BEFORE ---
   2 <script nonce="f59gFPKugO6+Xvi71Q0EWw==">
   1 <script type="application/ld+json">
   1 <script>                                    ← setInitialMode(…), no nonce

--- AFTER ---
   2 <script nonce="rMOmRoy3Aymy6Pbz8moJ4A==">
   1 <script type="application/ld+json">
```

(`application/ld+json` is a data block — the HTML spec never executes it, so CSP never checks it. Verified on `/`, `/events`, `/login`, `/organizations`.)

Real Chromium console against the two builds side by side:

```
BEFORE (unfixed preview)  csp-violations=1
    Executing inline script violates the following Content Security Policy directive 'script-src 'self' 'nonce-idTh8tvLr2IxELsfy3eOQA==''. Either the 'uns…
AFTER  (fixed preview)    csp-violations=0
```

## No FOUC regression

The existing `#440` specs plus the new `#816` guard, run against a production build+preview on chromium / firefox / webkit:

```
✓ production CSP › ships no nonce-less executable inline script (#816)
✓ production CSP › document responses carry a nonce-based script-src without unsafe-inline
✓ production CSP › runtime API origin is appended to connect-src / img-src / media-src (#396)
✓ dark-mode anti-FOUC (#440) › serves a nonced anti-FOUC script that the CSP allow-lists
✓ dark-mode anti-FOUC (#440) › applies the dark class before hydration for a dark-mode user
✓ dark-mode anti-FOUC (#440) › does not force dark on an explicit light-mode user

  18 passed (7.9s)
```

The new guard is not vacuous — it **failed** when first run against a stale preview server built from `main`, and passes on this branch.

## Quality gate

```
$ make fix && make check
… ✓ Type gate is armed (deliberate errors in .ts and .svelte both caught)
… ✅ No new hardcoded user-facing strings (i18n).
… ✅ All files are within line limits (Svelte: 750, TS/JS: 500).
… ✓ No SSR access-token leaks
… ✓ image-alt audit clean
make check exit code: 0
```

Closes #816

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved protection against content security policy violations by ensuring inline scripts use a valid security nonce.
  - Prevented flashes of unstyled content during page loading with a consistent synchronous startup script.
  - Preserved appearance-mode detection and switching while eliminating duplicate automatic script injection.

- **Tests**
  - Added regression coverage to detect executable inline scripts that lack the required security nonce.

### DIFF
--- a/src/app.html
+++ b/src/app.html
@@ -27,11 +27,15 @@
 		-->
 		<!--
 			Anti-FOUC: apply the dark class before first paint so dark-mode users
-			never see a light flash on SSR loads (#440). mode-watcher injects an
-			equivalent <svelte:head> snippet, but it carries no nonce and is blocked
-			by our production CSP (script-src 'self'); this script is nonced via
-			%sveltekit.nonce% so it runs. Must stay synchronous (no defer/async) and
-			mirror mode-watcher's storage key ('mode-watcher-mode') and dark class.
+			never see a light flash on SSR loads (#440). This is the ONLY anti-FOUC
+			script: mode-watcher used to inject an equivalent <svelte:head> snippet,
+			but it carried no nonce, was blocked by our production CSP
+			(script-src 'self' 'nonce-…') and logged a violation on every page load,
+			so `<ModeWatcher disableHeadScriptInjection />` in src/routes/+layout.svelte
+			now turns it off (#816). This script is nonced via %sveltekit.nonce% so it
+			runs. Must stay synchronous (no defer/async) and mirror mode-watcher's
+			storage key ('mode-watcher-mode') and dark class — mode-watcher no longer
+			has a fallback to cover a drift here.
 
 			Embed documents (/embed/*) with an explicit ?theme=light|dark carry
 			data-theme-locked and are skipped: their theme is already applied to
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -298,7 +298,19 @@
 	-->
 	{@render children()}
 {:else}
-	<ModeWatcher />
+	<!--
+		`disableHeadScriptInjection` (#816): mode-watcher's own FOUC snippet is
+		injected via {@html} as a BARE inline <script>. Its `nonce` prop can't help
+		us — SvelteKit's per-response nonce is only interpolated into app.html
+		(`%sveltekit.nonce%`) and is never exposed to components or load functions —
+		so under our production CSP (`script-src 'self' 'nonce-…'`) the browser
+		refused it and logged a violation on EVERY page load. The snippet never ran
+		anyway: the equivalent nonced script in src/app.html does the same job (same
+		'mode-watcher-mode' storage key, same `dark` class, same colorScheme), so
+		turning the injection off changes nothing but the console noise. Keep the two
+		in sync — app.html is now the ONLY anti-FOUC script.
+	-->
+	<ModeWatcher disableHeadScriptInjection />
 	<QueryClientProvider client={queryClient}>
 		<Toaster richColors position="top-right" />
 		<ImpersonationBanner />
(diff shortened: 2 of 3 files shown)


biagiodistefano:
@coderabbitai review

biagiodistefano:
@coderabbitai full review

biagiodistefano:
Addressed in fc745c3d — the guard now requires a real `src`/`nonce` attribute with a non-empty quoted value (`data-src=`/`data-nonce=`/`nonce=""` no longer count), keeping the ld+json exemption.

PR #822, opened 2026-08-07
fix(checkout): treat /api/me/billing 404 as an empty state, not an error

### LINKED ISSUE #817: checkout: /api/me/billing 404 surfaces as a console error on every paid-checkout dialog
Found during the #812 pre-merge smoke pass.

**Symptom:** opening the paid-ticket checkout dialog fires `GET /api/me/billing`, which returns 404 for any user without a billing record — semantically "nothing saved yet", but it lands as a red console error on every first-time purchase.

**Fix sketch:** treat 404 from this endpoint as an expected empty state in the query (e.g. return null instead of throwing, `retry: false`), so the console stays clean. Alternatively coordinate with BE to return 200 + empty payload — but the FE-side handling is the smaller change and needed regardless.

biagiodistefano:
## Symptom

Opening the paid-ticket checkout dialog fired `GET /api/me/billing` on mount for every authenticated buyer. The endpoint 404s for anyone who has never saved billing details (BE: `get_object_or_404(UserBillingProfile, user=...)`) — semantically "nothing saved yet" — so a first-time purchase logged a red failed-request line on every dialog open.

## What was already true

The queryFn in `CheckoutBillingSection.svelte` already returned `null` on 404 with `retry: false`, and the generated client returns `{ error, response }` rather than throwing when `throwOnError` is off. So nothing ever reached a TanStack error state, and no `console.error` in the app (or in the API client interceptors, or the app `QueryClient`) fires for a 404. The red line is the **browser's own** network log for the 404 response — JS handling cannot suppress it; only not making the request can.

## Fix

The billing profile is read by exactly one thing: the prefill that runs when the user ticks "request invoice". So the query is now gated on the section being open as well as on auth:

- A plain paid-checkout dialog now makes **no** `/api/me/billing` request at all — the symptom ("on every paid-checkout dialog") is gone.
- The 404 only happens for a user who actually asked for an invoice, and there it stays what it always was: an empty form, no error state, no retry.
- Prefill still lands (guarded by `if (!field)`, so it never clobbers anything typed in the meantime); the only behavioural change is that it arrives a request later than before, after the tick, instead of being pre-warmed on mount.

The three `enabled` operands are read unconditionally in a `$derived.by` — `&&` evaluated inside the tracked options getter would leave the later ones unread, and therefore untracked, freezing `enabled` at its initial value (the tracked-props freeze from #629).

The 404-as-null handling itself is unchanged; the comments now explain why it is there so it does not get "simplified" away.

## Consumers checked

All three callers of `userbillingGetBillingProfile` share the `['user-billing-profile']` query key; none uses the error state as its empty-UI branch:

| Consumer | 404 handling | Change |
| --- | --- | --- |
| `tickets/CheckoutBillingSection.svelte` | 404 → `null`, all other errors → `null` (best-effort prefill), `retry: false` | gated on section open |
| `billing/BillingProfileForm.svelte` (account settings) | 404 → `null`, other errors throw, `retry: false`; empty state via `hasBillingProfile` | none — the request is expected on a page dedicated to billing |
| `(auth)/account/referral/+page.svelte` | same shape as above | none — same reason |

## Tests

`CheckoutBillingSection.test.ts` gains three cases: no request while collapsed (then exactly one after opening), a 404 yielding a blank form with no `console.error`, and — under a `retry: 3` query client — a single call landing in `status: 'success'` with `null` data.

Canary: reverting the `enabled` gate makes the first new test fail, so it is decisive rather than vacuous.

## Follow-up (not in scope)

For a user who *does* tick "request invoice" with no saved profile, the browser still logs that one 404. Removing it entirely needs the BE to answer `200` + empty payload, as the issue notes.

## Verification

```
$ pnpm vitest run src/lib/components/tickets/CheckoutBillingSection.test.ts src/lib/components/billing/BillingProfileForm.test.ts
 Test Files  2 passed (2)
      Tests  33 passed (33)

$ make fix && make check   # EXIT=0
All matched files use Prettier code style!
pnpm eslint . --max-warnings 0
✓ All validations passed!            (i18n)
COMPLETED 6176 FILES 0 ERRORS 381 WARNINGS   (svelte-check; warnings = pre-existing baseline)
✓ No SSR access-token leaks
✓ image-alt audit clean
```

Closes #817

### DIFF
--- a/src/lib/components/tickets/CheckoutBillingSection.svelte
+++ b/src/lib/components/tickets/CheckoutBillingSection.svelte
@@ -69,7 +69,21 @@
 	// Form validation
 	let billingNameError = $state('');
 
-	// 4. Query: pre-fill from billing profile when authenticated
+	// 4. Query: pre-fill from billing profile when the user asks for an invoice.
+	// Gated on `isOpen` as well: the profile is read ONLY by the prefill below, so
+	// firing on mount spent a request on every paid-checkout dialog — and for the
+	// many users with nothing saved the endpoint answers 404, which the browser
+	// still logs as a failed request even though we handle it (#817).
+	// All three operands are read unconditionally: `&&` evaluated inside the
+	// tracked options getter would leave the later ones unread — and therefore
+	// untracked — freezing `enabled` at its initial value.
+	const shouldLoadBillingProfile = $derived.by(() => {
+		const authenticated = isAuthenticated;
+		const hasToken = !!authToken;
+		const sectionOpen = isOpen;
+		return authenticated && hasToken && sectionOpen;
+	});
+
 	const billingProfileQuery = createQuery<UserBillingProfileSchema | null>(() => ({
 		queryKey: ['user-billing-profile'],
 		queryFn: async () => {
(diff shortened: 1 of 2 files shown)


biagiodistefano:
@coderabbitai review

biagiodistefano:
@coderabbitai full review

Issue #814, opened 2026-08-07
i18n: "RSVP by in 6 days" — doubled preposition when composing rsvpBy with the relative phrase; no absolute deadline shown


biagiodistefano:
Follow-up to #812 (found in its pre-merge smoke pass; pre-existing composition, #812 only localized the relative phrase itself).

**Symptom:** the public event page renders the RSVP deadline as `RSVP by in 6 days` — `eventQuickInfo.rsvpBy` ("RSVP by {deadline}") is fed the output of `getRSVPDeadlineRelative()`, which already carries its own directional word ("in 6 days" / "in 6 Tagen" / "em 6 dias"). Every locale gets the same doubled-preposition shape: de `Anmeldung bis in 6 Tagen`, pt `Inscrição até em 6 dias`.

**Also:** the deadline is only ever shown relatively — no absolute date/time anywhere on the public page (no tooltip/title either), so the exact cutoff is not discoverable.

**Fix sketch:** either drop the prefix and let the relative phrase stand alone under the "RSVP deadline" label (EventDetails already does exactly that and reads fine), or reword the message keys per-locale to take a bare duration. Consider adding the absolute deadline (e.g. `formatDateTime`) as secondary text or `title`.

Surfaces: `EventQuickInfo.svelte` (rsvpBy composition), message keys `eventQuickInfo.rsvpBy` across the six locales.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01JDXmzqVQzDQTBwT98KL9Wf

PR #823, opened 2026-08-07
fix(dates): consistent helpers, machine-readable card dates, tz context on dashboard lists

### LINKED ISSUE #818: dates: display consistency follow-ups from the #812 smoke pass
Grouped minor findings from the #812 pre-merge smoke pass — all cosmetic/semantic, none functional:

1. **Ticket modals use two different date helpers**: the ticket list and the *active*-ticket modal render `Sat, Sep 26 • 8:29 PM` (formatEventDate) while the *pending*-ticket modal renders `Sep 26, 2026, 8:29 PM` (formatDateTime). Pick one per surface.
2. **Event cards render dates in plain spans** — only the event detail page and the RSVPs list use `<time datetime="…">`. List surfaces (discovery grid, dashboard grids) lose machine-readable dates (SEO/structured-data relevant).
3. **Timezone context differs between surfaces**: the event detail page shows event-local times with the "Times shown in <place> (GMT±X)" disclosure, but `/dashboard/rsvps` shows the same event in viewer-local time with no tz indicator. Both are correct conversions; a user comparing the two sees different clock times with no explanation. Consider the tz label (or the event's zone) on dashboard lists too.

biagiodistefano:
Grouped display-consistency follow-ups from the #812 smoke pass. All three findings from #818, plus the verification below.

## 1. Ticket modals used two different date helpers → one helper per surface

**Chosen: `formatEventDate`, in the event's timezone.**

The split wasn't pending-vs-active by design — it was per *call site*. `TicketListCard` passes `formatEventDateRange(…)` into `MyTicketModal` (hence `Sat, Sep 26 • 8:29 PM`), while the event page passed `formatDateTime(event.start)` into both `MyTicket` and `MyTicketModal` (hence `Sep 26, 2026, 8:29 PM` on the pending ticket). Both event-page call sites now use `formatEventDate(event.start, event.timezone)`.

- The year isn't load-bearing here: the ticket is always shown next to the event it belongs to, and the list surface has never shown one.
- Passing `event.timezone` fixes a second, quieter inconsistency on that page: everything else there (header, details, schedule) renders **event-local** times, while the ticket card/modal was rendering **viewer-local** — the exact bug class this issue is about, inside a single page.
- The tz abbreviation is deliberately **not** suppressed (no `withAbbreviation: false`): a ticket is read detached from the page's shared "Times shown in …" note — in a modal, at the door.

## 2. Event cards render dates in plain spans → `<time datetime>`

Wrapped the date text in `<time datetime="<ISO>">` (ISO 8601 in the attribute, existing helper output as the visible content) on every list/card surface:

`EventCard` (discovery grid, org page, series page, dashboard grid, embeds) · `RSVPCard` · `TicketListCard` · `InvitationCard` · `AdminEventCard`.

Created/purchased/invited dates in the card footers got the same treatment. **No visual change** — `<time>` is inline like the `<span>` it replaced, and every one of these sits in a flex row, so `truncate` behaves identically.

## 3. Timezone context differs between surfaces → name the viewer's zone on dashboard lists

**Chosen: a compact viewer-timezone note, not event-local rendering.**

Rendering event-local times on the dashboard is not currently possible: `/api/dashboard/rsvps` and `/api/dashboard/tickets` serialize the event as `MinimalEventSchema`, which has **no `timezone` field** (the event page gets it from `EventDetailSchema`). So converting to the event's zone there would need a backend change.

The smallest honest fix is to say which zone the times *are* in:

- `formatViewerTimezoneLabel()` in `date.ts` — the counterpart to the existing `formatEventTimezoneLabel()`, returning `"CEST"` / `"GMT+2"`. No place name: the viewer's location is not news to the viewer; the offset is what makes the two surfaces comparable.
- `ViewerTimezoneNote.svelte` — the twin of `EventTimezoneNote.svelte` (same `Globe` + `text-xs text-muted-foreground` treatment), rendered once above the grid on `/dashboard/rsvps` and its twin `/dashboard/tickets`.
- **Browser-only by construction**: on the server `resolvedOptions()` reports the *server's* zone, so an SSR render would state a timezone that isn't the viewer's. Rendering nothing there is the honest fallback (and the lists are client-only in practice — the query is gated on the in-memory access token).

New message key `eventTime.shownInYours` in **all six locales** ("Times shown in your time zone ({label})").

A follow-up worth filing on the backend: adding `timezone` to `MinimalEventSchema` would let the dashboard lists show event-local times and drop the note entirely.

Not touched, per the parallel work on #814: `EventQuickInfo.svelte` and the `eventQuickInfo.*` keys.

## Verification

```
$ pnpm vitest run
 Test Files  221 passed (221)
      Tests  2616 passed | 1 todo (2617)

$ make fix && make check
… COMPLETED 6177 FILES 0 ERRORS 381 WARNINGS 55 FILES_WITH_PROBLEMS   # warnings are the pre-existing baseline
✓ All validations passed!   # i18n: 6 languages, 100% completion, placeholders consistent
make check exit: 0
```

`svelte-autofixer` clean on the new component. Three unit tests added for `formatViewerTimezoneLabel` (`src/lib/utils/date.test.ts`), TZ-agnostic so they hold on any machine.

Both `date.ts` (568/570) and the event page (748/750) were within a few lines of their length caps; the change is written to fit under them rather than raise them.

Closes #818

Co-Authored-By: Claude <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Added localized notices identifying when event, RSVP, and ticket times are shown in the viewer’s timezone.
  * Improved event-time displays to consistently use the event’s timezone where applicable.
  * Added machine-readable date and time information for improved accessibility and calendar integration.

* **Bug Fixes**
  * Improved timezone labels across supported languages and ensured they remain consistent during daylight-saving changes.

### DIFF
--- a/messages/de.json
+++ b/messages/de.json
@@ -1600,7 +1600,8 @@
 		"failedToCopyLink": "Link konnte nicht kopiert werden"
 	},
 	"eventTime": {
-		"shownIn": "Zeiten werden in {label} angezeigt"
+		"shownIn": "Zeiten werden in {label} angezeigt",
+		"shownInYours": "Zeiten werden in deiner Zeitzone angezeigt ({label})"
 	},
 	"eventInvitationsAdmin": {
 		"backToEvents": "Zurück zu den Events",
--- a/messages/en.json
+++ b/messages/en.json
@@ -1600,7 +1600,8 @@
 		"failedToCopyLink": "Failed to copy link"
 	},
 	"eventTime": {
-		"shownIn": "Times shown in {label}"
+		"shownIn": "Times shown in {label}",
+		"shownInYours": "Times shown in your time zone ({label})"
 	},
 	"eventInvitationsAdmin": {
 		"backToEvents": "Back to Events",
--- a/messages/es.json
+++ b/messages/es.json
@@ -1600,7 +1600,8 @@
 		"failedToCopyLink": "No se ha podido copiar el enlace"
 	},
 	"eventTime": {
-		"shownIn": "Horas mostradas en {label}"
+		"shownIn": "Horas mostradas en {label}",
+		"shownInYours": "Horas mostradas en tu zona horaria ({label})"
 	},
 	"eventInvitationsAdmin": {
 		"backToEvents": "Volver a eventos",
--- a/messages/fr.json
+++ b/messages/fr.json
@@ -1486,7 +1486,8 @@
 		"failedToCopyLink": "Échec de la copie du lien"
 	},
 	"eventTime": {
-		"shownIn": "Horaires affichés en {label}"
+		"shownIn": "Horaires affichés en {label}",
+		"shownInYours": "Horaires affichés dans votre fuseau horaire ({label})"
 	},
 	"eventInvitationsAdmin": {
 		"backToEvents": "Retour aux événements",
--- a/messages/it.json
+++ b/messages/it.json
@@ -1600,7 +1600,8 @@
 		"failedToCopyLink": "Impossibile copiare il link"
 	},
 	"eventTime": {
-		"shownIn": "Orari mostrati in {label}"
+		"shownIn": "Orari mostrati in {label}",
+		"shownInYours": "Orari mostrati nel tuo fuso orario ({label})"
 	},
 	"eventInvitationsAdmin": {
 		"backToEvents": "Torna agli eventi",
--- a/messages/pt.json
+++ b/messages/pt.json
@@ -1600,7 +1600,8 @@
 		"failedToCopyLink": "Falha ao copiar o link"
 	},
 	"eventTime": {
-		"shownIn": "Horários mostrados em {label}"
+		"shownIn": "Horários mostrados em {label}",
+		"shownInYours": "Horários mostrados no seu fuso horário ({label})"
 	},
 	"eventInvitationsAdmin": {
 		"backToEvents": "Voltar aos eventos",
--- a/src/lib/components/events/EventCard.svelte
+++ b/src/lib/components/events/EventCard.svelte
@@ -225,10 +225,12 @@
 				variant === 'compact' && 'gap-1.5 border-t-0 pt-0 md:border-t md:pt-3'
 			)}
 		>
-			<!-- Date & Time -->
+			<!-- Date & Time. <time datetime> so the machine-readable instant survives
+			     on list surfaces too (crawlers, structured data) — the visible text
+			     stays the localized, event-timezone rendering. -->
 			<div class="flex items-center gap-2 text-sm">
 				<Calendar class="h-4 w-4 shrink-0 text-muted-foreground" aria-hidden="true" />
-				<span class="truncate">{formattedDate}</span>
+				<time datetime={event.start} class="truncate">{formattedDate}</time>
 			</div>
 
 			<!-- Location -->
--- a/src/lib/components/events/ViewerTimezoneNote.svelte
+++ b/src/lib/components/events/ViewerTimezoneNote.svelte
@@ -1,0 +1,32 @@
+<script lang="ts">
+	import * as m from '$lib/paraglide/messages.js';
+	import { browser } from '$app/environment';
+	import { Globe } from '@lucide/svelte';
+	import { formatViewerTimezoneLabel } from '$lib/utils/date';
+
+	interface Props {
+		/** ISO instant used to resolve the (DST-aware) offset; defaults to now. */
+		reference?: string;
+		/** Optional extra classes for layout tuning by the host. */
+		class?: string;
+	}
+
+	const { reference, class: className = '' }: Props = $props();
+
+	// Twin of EventTimezoneNote, for lists that can only render viewer-local
+	// times (the dashboard payloads carry no event timezone). Naming the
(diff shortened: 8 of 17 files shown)


biagiodistefano:
### Code review

Found 1 issue:

1. `InvitationCard` renders viewer-local times with a comment justifying it as unavoidable — but `invitation.event` is `EventInListSchema`, which carries a required `timezone: string` (unlike the `MinimalEventSchema` used by the RSVP/ticket dashboards this PR adds `ViewerTimezoneNote` to). The timezone is available and unused: `eventDate` calls `formatEventDateRange(invitation.event.start, invitation.event.start)` with no timezone, while this same PR updates `EventCard` (same schema) to pass `event.timezone`. Invitations therefore still show undisclosed viewer-local times that differ from the event page — the exact failure mode issue #818 targets — and the invitations dashboard got no `ViewerTimezoneNote` disclosure either. Either pass `invitation.event.timezone` to the formatter (matching `EventCard`) or correct the comment and add the disclosure note.

https://github.com/letsrevel/revel-frontend/blob/30deedb47fbcdd7b5f721a9ae6a67c2abf3e4225/src/lib/components/invitations/InvitationCard.svelte#L26-L30

https://github.com/letsrevel/revel-frontend/blob/30deedb47fbcdd7b5f721a9ae6a67c2abf3e4225/src/lib/components/invitations/InvitationCard.svelte#L104-L108

Schema evidence: https://github.com/letsrevel/revel-frontend/blob/30deedb47fbcdd7b5f721a9ae6a67c2abf3e4225/src/lib/api/generated/types.gen.ts#L1450-L1452

biagiodistefano:
Review findings addressed in d993364a. All three were re-verified against `src/lib/api/generated/types.gen.ts` before fixing; all three held.

**Finding 1 — InvitationCard viewer-local (held).** Confirmed: `MyEventInvitationSchema.event` is an `EventInListSchema`, which declares a required `timezone: string`. The "dashboard payloads carry no event timezone" justification is real, but it applies to `UserRsvpSchema`/`UserTicketSchema` (both `MinimalEventSchema`, which genuinely has no `timezone` field) — not to this card. Now renders event-local:

```
formatEventDate(invitation.event.start, invitation.event.timezone)
```

The old call passed `start` twice to `formatEventDateRange`, so it was always a single instant; the single-instant helper states that directly, and matches `EventCard` on the same schema. The misleading comment is corrected.

**Finding 2 — AdminEventCard viewer-local (held).** Same root cause: its `event` prop is also `EventInListSchema` with required `timezone`. Verified both siblings named in the finding — `EventCard.svelte:50` and `org/[slug]/admin/tickets/+page.svelte:73` — and both use `formatEventDate(event.start, event.timezone)`.

Choice, stated explicitly: I adopted the sibling **timezone-passing** convention but kept `formatDateTime` as the helper — `formatDateTime(event.start, event.timezone)`. Rationale: `date.ts` documents `formatDateTime` as the admin-pages/lists format, it is the Studio-volume convention, and it keeps the **year**, which an admin list mixing drafts, open and past events needs (`formatEventDate` drops it). The defect was the missing zone, not the helper; swapping the helper too would have been an unrequested visual change to the admin list.

**Finding 3 — ViewerTimezoneNote DST-sensitive label (held).** Confirmed both call sites (`dashboard/rsvps`, `dashboard/tickets`) omit `reference`, so the abbreviation resolved from *now*. Measured the contradiction rather than assuming it — for a single viewer zone:

```
OLD (offset-based): winter=GMT+1  summer=GMT+2
NEW (zone name)   : Vienna (both)
```

So a note rendered in January over a list containing July events was literally mislabelling those rows.

Judgment call: **I rejected passing the soonest displayed event's start.** It is still wrong for every row on the far side of a transition, and it makes the label flicker as the user filters or paginates — a page-level note cannot carry a per-row offset, so it should not claim one. Instead `formatViewerTimezoneLabel()` now returns a DST-invariant zone **name** humanized from the viewer's IANA zone (`"Vienna"`), which is true for every row on the page and still supplies what the note exists for: *which zone these clock times belong to*, so they can be told apart from the event page's event-local ones. `reference` is dropped from both the helper and the component (it existed only to resolve the offset being removed), so there are **no call-site changes and no copy changes** — the existing key renders "Times shown in your time zone (Vienna)". The reasoning is documented in both the component and the helper docstring.

Nothing was rejected on re-verification; the only rejected item is the *implementation option* above for Finding 3.

**Verification**

- `pnpm vitest run` — `221 passed (221)` files, `2617 passed | 1 todo`.
- `make fix` then `make check` — **exit 0** (`0 ERRORS`, type gate confirmed armed, i18n clean, file lengths clean, no SSR token leaks, image-alt clean).
- The three `formatViewerTimezoneLabel` tests are rewritten for the new contract, plus a new **DST-invariance regression test** (same label either side of a transition) that fails against the old offset-based implementation.
- Note: `make check` initially failed `file-length` — `date.ts` hit 576 against its sanctioned 570 cap. I tightened the new docstring back to 570 rather than raising the override, since the override map was out of scope.

No `messages/*.json` changes were needed, so no catalog grep or recompile was required beyond the initial `pnpm paraglide:compile`. `svelte-autofixer` reports no issues on all three components.

Issue #101, opened 2025-10-21
Phase 2.2: Add resources section to event edit page


biagiodistefano:
Integrate resources into event edit wizard allowing assignment from event context.

**Estimated Effort:** 4-6 hours  
**Size:** M

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #824, opened 2026-08-07
fix(announcements): searchable paginated event picker in the composer

### LINKED ISSUE #815: announcements: event picker caps at 100 options with no search — later events unselectable
Found during the #812 pre-merge smoke pass.

**Symptom:** the announcement composer's event picker lists at most 100 events and offers no search/pagination. In an org with ~100+ events (dev seed + E2E data, but any large org hits this), events beyond the cap — e.g. "Summer Sunset Music Festival" — cannot be selected at all, so no announcement can target their attendees.

**Fix sketch:** searchable combobox backed by the paginated events endpoint (typeahead), or at minimum raise the page size and sort upcoming-first so active events are reachable.

biagiodistefano:
## Symptom

The announcement composer's event picker was a plain `<select>` populated by one `page_size: 100` call to `/api/events/`. In an org with 100+ events (dev seed, E2E data, or any real large org), every event past the cap was simply missing from the dropdown — so no announcement could ever be sent to those events' attendees. Found during the #812 pre-merge smoke pass ("Summer Sunset Music Festival" was unselectable).

## Approach

New `src/lib/components/announcements/EventCombobox.svelte` — a debounced (250 ms) typeahead over the **same** paginated endpoint, wired through TanStack Query.

**API params used** (all present in the generated client, `EventpublicdiscoveryListEventsDbaec952Data` in `src/lib/api/generated/types.gen.ts:21586`, so none of them are silently dropped by Django Ninja):

| param | value |
| --- | --- |
| `organization_slug` | the composer's org |
| `search` | debounced query (backend `@searching` over name / description / series / org / tags) |
| `include_past` | driven by the existing "Include past events" checkbox |
| `order_by` | `'start'` (upcoming-first) — flips to `'-start'` when past events are included, since ascending `start` would bury every live event under years of history |
| `page_size` | `20` |

**Preserving the selection.** The label for the selected event is owned by the composer, not by the result page: it comes from `announcement.event_name` when editing, and from a new `preSelectedEventName` prop at the event-page call site. So the field stays readable even when the selected event is nowhere near the rows currently loaded.

**Simplification.** Fetching now lives inside the picker, so the three call sites that existed only to hand the modal their own capped event list (`admin/announcements/+page.svelte`, `admin/+page.svelte`, `EventManageSection.svelte`) drop that query and the `events` / `eventsLoading` / `includePastEvents` / `onIncludePastChange` prop plumbing entirely.

## Accessibility

ARIA 1.2 editable-combobox pattern with virtual focus:

- `role="combobox"` + `aria-autocomplete="list"` + `aria-expanded` + `aria-activedescendant`; `aria-controls` is set **only** while the listbox is mounted (no dangling IDREF when collapsed).
- Keyboard: Arrow Up/Down (open + wrap), Home/End, Enter to pick — swallowed so it can't submit the surrounding draft form — Escape to close the list, `stopPropagation()`'d so it doesn't take the whole bits-ui dialog down with it, and Tab to dismiss.
- The clear button is a real `<button>` named by `sr-only` text next to an `aria-hidden` icon, 36×36.
- sr-only `role="status"` result count; the target validation error is now given an id and wired into the field's `aria-describedby`, so an invalid picker announces *why*.
- Mobile-first: full-width field, list capped at `min(15rem, 40vh)` so it survives a phone with the software keyboard up, and the picker column is no longer indented below `sm`.

Audited by the accessibility-checker subagent; all three defects it raised (dangling `aria-controls`, a stale blur timer that slammed the reopened list shut after "clear", and the missing error association) are fixed in this branch.

## i18n

5 new keys under `announcements.form.*` (`eventSearchPlaceholder`, `eventSearchHint`, `clearEventSelection`, `noEventsMatch`, and the CLDR-plural `eventResultsCount`) added to **all six** locales; `pnpm paraglide:compile` run.

## Verification

```
$ make fix && make check
All matched files use Prettier code style!
pnpm eslint . --max-warnings 0        # clean
COMPLETED 6177 FILES 0 ERRORS 381 WARNINGS   # warnings are the pre-existing baseline; none in touched files
✓ No SSR access-token leaks
✓ image-alt audit clean
EXIT=0

$ pnpm vitest run
Test Files  222 passed (222)
     Tests  2621 passed | 1 todo (2622)
```

8 new component tests in `EventCombobox.test.ts` cover the ARIA contract, both `order_by` modes, keyboard selection, clearing, the "no match" state, label preservation for an off-page selection — and, centrally, **reaching event #101 by searching for it** over a 101-event catalogue whose unsearched page returns only the first 20. The blur-timer regression test was verified non-vacuous by reverting the fix and watching it fail.

Not run: E2E (the announcements journey spec never touches the event picker, so no spec changes were needed).

Closes #815

Co-Authored-By: Claude <noreply@anthropic.com>


biagiodistefano:
### Code review

Found 1 issue:

1. Enter is swallowed for as long as the event field has focus, blocking keyboard form submission. The listbox opens on mere focus (`onfocus={openList}`), and the `Enter` case does `event.preventDefault()` unconditionally while `open` is true — but only *acts* when an option is highlighted (`activeIndex >= 0`). With the list open and nothing highlighted (the state you land in just by focusing or tabbing into the field), Enter neither submits the enclosing `<form>` (which supports implicit submit via its single `type="submit"` button, and did with the old `<select>`) nor selects, closes, or gives any feedback. A keyboard user must press Escape first or reach for the mouse. Suggested fix: when `activeIndex < 0`, either let Enter through to the form, or have it close the list and then let a second Enter submit — swallow it only when a highlight will actually be committed.

https://github.com/letsrevel/revel-frontend/blob/5744481b930326b7cef59ced957515623b6768da/src/lib/components/announcements/EventCombobox.svelte#L205-L213

https://github.com/letsrevel/revel-frontend/blob/5744481b930326b7cef59ced957515623b6768da/src/lib/components/announcements/EventCombobox.svelte#L248-L251

biagiodistefano:
Both review findings re-verified against the code and fixed in 7899e1c0.

**Finding 1 — dead Enter key (held).** Confirmed: `onfocus={openList}` means the listbox is open the moment the field is focused, and the `Enter` case did `event.preventDefault()` for the whole `open` state while only acting on `activeIndex >= 0`. Focus + Enter with nothing highlighted was a no-op — no selection, and no implicit submit of the composer `<form>` (which does have a `<button type="submit">`, so the old `<select>` form did submit).

Chosen behaviour: **swallow Enter only when a highlight will be committed; otherwise let it through untouched to the form** (no `preventDefault`, and the list is deliberately *not* closed either — a failed validation then leaves the user's search text and results intact; Escape stays the explicit way to close). Rationale: the [ARIA APG combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/) defines Enter for an editable combobox solely for the case where "an autocomplete suggestion is selected in the popup" — with nothing selected the key has no combobox meaning and should keep its default action.

Verified the safety question raised in the review: leftover half-typed search text cannot leak into a submit. `query` is display-only; the selection lives in the composer's `selectedEventId`/`selectedEventName`, and `getFormData()` reads `selectedEventId`.

**Finding 2 — `orgHasNoEvents` ignored `includePast` (held).** Confirmed against the backend: `EventPublicDiscoveryController.list_events` does `params.next_events = not include_past`, so with the default `includePast=false` an org whose events are *all* in the past returns zero rows for an unsearched query — and the picker claimed "No events available / Create an event first, or enable past events to see older ones." Factually wrong, and its own companion string gave the game away.

Now split on the axis the result can actually prove:
- `includePast` **true**, no search, zero rows → `announcements.form.noEvents` ("No events available"). Its description also dropped the now-contradictory "or enable past events" clause, since that branch only renders when past events *are* included.
- `includePast` **false**, no search, zero rows → new `announcements.form.noUpcomingEvents` / `.noUpcomingEventsDescription` ("No upcoming events" / "Enable past events to search older ones").

New keys added to all six locales (en/it/de/fr/es/pt), matching neighbouring register (informal *du*/*tu*/*tú*, de kept gender-neutral via "Teilnehmenden"); `pnpm paraglide:compile` run, `validate-translations` reports 8005 aligned keys per locale.

Nothing was rejected on re-verification — both findings reproduced exactly as described.

**a11y contract kept intact**: `aria-activedescendant`/`aria-controls` gating, the Escape `stopPropagation`, and the sr-only status region are untouched. `svelte-autofixer` reports 0 issues on the component.

**Tests.** Both fixes carry a test that I verified *fails* against the pre-fix component (reverted the two hunks, re-ran: `2 failed | 10 passed`, "expected vi.fn() to be called 1 times, but got 0 times" for the Enter case), plus companions:
- `lets Enter submit the enclosing form when no option is highlighted` — mounts the combobox inside a real `<form>` with a submit button (user-event only performs implicit submission when the keydown was not swallowed)
- `swallows Enter instead of submitting when an option is highlighted`
- `offers past events instead of claiming the org has none when only upcoming are searched`
- `claims the org has no events only once past events are included`

**Verification**: `pnpm vitest run src/lib/components/announcements/EventCombobox.test.ts` → 12/12 passed · full `pnpm vitest run` → 222 files, 2625 passed · `make fix` + `make check` → exit 0 (type gate confirmed armed).

PR #826, opened 2026-08-09
fix(date): count RSVP deadlines in calendar days, unify relative-time formatting

florimond-collette:
## Problem

`getRSVPDeadlineRelative` measured deadlines in elapsed 24-hour blocks: `Math.floor(diffMs / 24h)`. A deadline that is on the **day after tomorrow** but less than 48 hours away rendered as **"in 1 day"** — e.g. now Sunday 23:00, deadline Tuesday 08:00 (33h) read "in 1 day" while every calendar surface next to it says Tuesday. Users plan against a calendar, not against 24-hour blocks.

## Change

**Calendar-day semantics from 24h out.** A new `calendarDaysBetween(from, to, timeZone?)` helper counts midnights crossed in a timezone (same cached en-CA `YYYY-MM-DD` formatter trick as the existing `isSameDayInZone`, differenced as UTC midnights — exact, DST-safe). Below 24 elapsed hours the hour unit still applies, even across a midnight ("in 23 hours", not "tomorrow"), so imminent deadlines keep their urgency.

**One engine instead of two near-duplicates.** `getRSVPDeadlineRelative` and `formatRelativeTime` are now thin wrappers over a shared `formatRelativeToNow(target, timeZone, minUnit)` that walks a year → month → day → hour → minute unit ladder. The single genuine difference between the callers became the `minUnit` parameter: `'minute'` clamps a seconds-away deadline to "in 1 minute" (an open RSVP must never read as already-current), `'second'` preserves `formatRelativeTime`'s "in 30 seconds" / "now".

**`numeric: 'auto'`.** Single-step values become the locale's idiom: "tomorrow" / "morgen" / "demain", "yesterday" — and German gets **"übermorgen"** for exactly the day-after-tomorrow case this PR fixes.

**DST edge handled by fall-through, not clamping.** On a 25-hour fall-back day (e.g. Europe/Vienna, 2026-10-25), an instant a full 24h away can cross *zero* midnights. "in 0 days" is nonsense and "tomorrow" would be false, so the day unit falls through to hours: "in 24 hours". Pinned by a test.

**Invalid input no longer throws (review finding).** An unparseable date string produced a `NaN` diff that reached `Intl.RelativeTimeFormat.format()`, which throws `RangeError` — crashing both exports (`getRSVPDeadlineRelative`'s passed-deadline guard doesn't catch `NaN`). The shared engine now checks `Number.isFinite(diffMs)` once and returns `""`, matching the `formatDateTimeReadback` convention for invalid input. Note: `getRSVPDeadlineRelative` returns `""` here, not `null` — `null` specifically means "deadline passed" per its contract.

**Review cleanups.** `SUPPORTED_UI_LANGUAGES` is now exported from `date.ts`, derived from `LOCALE_MAP`, and the textual-month invariant test iterates it instead of a hand-synced hardcoded list — adding a language to the map automatically extends the checks. `freezeTime` in the test helpers gained its `void` return annotation.

## Behavior changes to be aware of

Every scenario is shown in both supported reference locales: `en` (en-US) and `de` (de-DE). All strings are actual `Intl.RelativeTimeFormat` output.

| Scenario | Locale | Before | After |
|---|---|---|---|
| Deadline 46h away, crossing 2 midnights | en | in 1 day | in 2 days |
| | de | in 1 Tag | **übermorgen** |
| Deadline 70h away, crossing 3 midnights | en | in 2 days | in 3 days |
| | de | in 2 Tagen | in 3 Tagen |
| Deadline exactly 24h away (1 midnight) | en | in 1 day | tomorrow |
| | de | in 1 Tag | morgen |
| Deadline 23h away, on the next calendar day | en | in 23 hours | in 23 hours *(unchanged — hours keep urgency below 24h)* |
| | de | in 23 Stunden | in 23 Stunden |
| Deadline 30 minutes away | en | in 30 minutes | in 30 minutes *(unchanged)* |
| | de | in 30 Minuten | in 30 Minuten |
| Deadline seconds away (still open) | en | in 1 minute | in 1 minute *(unchanged clamp)* |
| | de | in 1 Minute | in 1 Minute |
| Deadline 45 days away | en | in 45 days | next month |
| | de | in 45 Tagen | nächsten Monat |
| DST fall-back: ≥24h away, same 25-hour local day | en | in 1 day | in 24 hours |
| | de | in 1 Tag | in 24 Stunden |
| Same instant pair, day boundary judged per timezone (26.5h away; 1 midnight in New York, 2 in Vienna) | en | in 1 day *(both zones)* | tomorrow (NY) / in 2 days (Vienna) |
| | de | in 1 Tag *(both zones)* | morgen (NY) / übermorgen (Vienna) |
| `formatRelativeTime`: instant 1 day ago | en | yesterday | yesterday *(unchanged — was already numeric:'auto')* |
| | de | gestern | gestern |
| `formatRelativeTime`: instant 46h ahead, 2 midnights | en | in 1 day | in 2 days *(inherits calendar counting)* |
| | de | in 1 Tag | übermorgen |
| Unparseable date string (both functions) | en | throws `RangeError` | `""` *(locale-independent)* |
| | de | throws `RangeError` | `""` |

- Both functions gain an optional `timeZone` (IANA) for judging the day boundary, consistent with the rest of `date.ts`. Default is the viewer's zone — the calendar an RSVP'ing viewer plans against. **Reviewers:** if a surface renders the deadline date in the *event's* timezone, pass that zone here too so the phrase and the date can't disagree across zones.
- `formatRelativeTime` call sites elsewhere in the app will show the new day-range phrasing; no signature change required.
- If deadlines should stay capped at the day unit (no "next month"), a `maxUnit` parameter on the engine is a two-line follow-up.
- Invalid input: `getRSVPDeadlineRelative` returns `""` (empty, falsy) rather than `null`, since `null` is reserved for "deadline passed".

## Tests

72 passing (63 existing + 9 new): calendar vs. elapsed counting (46h → "in 2 days"), hours kept below 24h across a midnight, exact-24h boundary → "tomorrow", the same instant pair judged per timezone (NY "tomorrow" vs. Vienna "in 2 days"), the German "übermorgen" idiom, the DST fall-back "in 24 hours" case, month escalation, shared calendar semantics in `formatRelativeTime`, and `""` (no throw) for unparseable input in both functions. Day-branch fixtures now pin `'UTC'` explicitly since output depends on the calendar zone; the supported-language invariant iterates the exported `SUPPORTED_UI_LANGUAGES` instead of a hardcoded copy.

## ⚠️ MUST READ — open UX question for the maintainer: sub-minute output

`formatRelativeTime` bottoms out at a seconds countdown: anything under a minute renders as "in 30 seconds" / "45 seconds ago" ("in 30 Sekunden", "tra 30 secondi", …). This precision is likely **bad or at least unneeded for the user's sense of perspective**: relative phrases exist to give a rough temporal orientation, and second-level counts read like a stopwatch — they're stale the moment they render (nothing re-renders per second), imply a precision the UI doesn't actually maintain, and add cognitive noise for a distinction ("in 12 seconds" vs. "in 40 seconds") no user acts on. **Consider clamping everything below one minute to `rtf.format(0, 'second')`**, which CLDR renders as the natural idiom in every supported locale: "now" / "jetzt" / "ora" / "maintenant" / "ahora" / "agora". This is a one-line change in `formatRelativeToNow` — replace the `Math.trunc(diffMs / 1000)` seconds branch with `rtf.format(0, 'second')` — and it would even let the `minUnit` parameter collapse away if applied unconditionally. **One deliberate exception:** keep `getRSVPDeadlineRelative` on its "in 1 minute" clamp rather than "now" — for a *deadline*, "now" reads as *already closed* (and `isRSVPClosed` treats the exact instant as closed), so a user at T−20s could wrongly conclude they've missed it; "in 1 minute" stays directional and honest. Not changed in this PR to keep the diff review-scoped — please decide and we'll follow up.

## Summary by CodeRabbit

* **New Features**
  * Added timezone-aware relative-time formatting for RSVP deadlines and general dates.
  * Added localized calendar terms such as “tomorrow.”
  * Added support for displaying relative times in months and years.

* **Bug Fixes**
  * Improved accuracy across daylight-saving changes and calendar-day boundaries.
  * Invalid dates are handled safely.
  * Passed RSVP deadlines now return no deadline message.
  * Relative-time calculations are more accurate for exact 24-hour periods.


biagiodistefano:
### Code review

Found 1 issue:

1. The new test file is named `date_test.ts` (underscore), but the vitest config only collects `src/**/*.{test,spec}.{js,ts}` — so none of the 521 lines of new tests (calendar-day counting, DST fall-back, timezone boundary, invalid-input guard) ever run. `pnpm test` and CI stay green while silently skipping the file; every other test file in the repo uses the `.test.ts` suffix, including the existing `src/lib/utils/date.test.ts`, which this PR leaves untouched and which still covers the old behavior. Renaming/merging into `date.test.ts` also fixes two side effects: the file-length gate currently fails on `date_test.ts` (521 > 500, because the script's `*.test.ts` exemption glob doesn't match either), and the `SUPPORTED_UI_LANGUAGES` export gets consumed by tests that actually execute.

Test file that never runs:
https://github.com/letsrevel/revel-frontend/blob/b9bde8a687ea24c9ec649ce544eeca27f4ee8189/src/lib/utils/date_test.ts#L1-L6

Vitest include pattern it fails to match:
https://github.com/letsrevel/revel-frontend/blob/b9bde8a687ea24c9ec649ce544eeca27f4ee8189/vite.config.ts#L17-L19

Note: CI's file-length gate will also flag `date.ts` itself (639 lines vs. its 570-line override in `scripts/check-file-length.sh`), so the file needs a trim/split or a re-justified cap bump — leaving that to CI since it reports it directly.

biagiodistefano (changes requested):
Requesting changes for the issue in my [review comment](https://github.com/letsrevel/revel-frontend/pull/826#issuecomment-5231185658): the new test file `src/lib/utils/date_test.ts` doesn't match vitest's `src/**/*.{test,spec}.{js,ts}` include pattern, so none of the new tests run — please rename/merge it into `src/lib/utils/date.test.ts`. That also clears the file-length gate failure on the test file; `date.ts` (639 lines vs. its 570-line cap in `scripts/check-file-length.sh`) still needs a trim/split or a re-justified cap bump for CI to go green.

biagiodistefano (dismissed):

Issue #894, opened 2026-09-07
feat(integrations): Eventbrite listings — organization Integrations page, event Listings tab, sales pause on both sides (FE for revel-backend#924)


biagiodistefano:
## Summary

Frontend for **letsrevel/revel-backend#924** (Eventbrite listing sync, phases 1–3). An organizer connects an Eventbrite account once per organization, then lists any public ticketed event on Eventbrite from Revel: first as a draft, published on request, kept up to date manually or automatically. Revel shows the tickets Eventbrite sold and can pause sales on either side. Existing Eventbrite events can be imported as Revel drafts.

Revel stays the place the event is edited. Eventbrite receives a copy. No attendee data crosses platforms, only sold counts.

Work lands on the integration branch `feature/894-integrations` (draft PR stays draft until every phase is in), one PR per phase into it, mirroring the backend.

### Decisions already taken

| Question | Decision |
| --- | --- |
| Where does the organization-level surface live | **New owner-only admin page** `/org/{slug}/admin/integrations`, in the admin nav and the dashboard quick actions, spliced in the same way as Billing. Not a section on the settings page, which is near the file-size cap and cannot host an account picker and an import picker. |
| Where does the event-level surface live | **Third editor tab "Listings"** (`edit?tab=listings`) next to Details and Ticketing, existing events only. Staff with `manage_event` can use it; only the owner can connect. |
| Naming | **"Integrations"** for the organization page (where an account is connected), **"Listings"** for the event tab (where this event appears elsewhere). Copy always names the platform ("Listed on Eventbrite"); the provider list itself comes from the API and is never hardcoded. |
| Tracking | One issue, phased checklist below. |

### Backend contract (branch `feature/integrations`)

Regenerate the client from that branch (`make dump-openapi` in revel-backend, then `pnpm generate:api` here). Everything below is in the generated types.

| Method | Path | Who | Returns |
| --- | --- | --- | --- |
| GET | `/organization-admin/{slug}/integrations` | owner | one `ConnectionSchema` per **enabled** provider; `status: null` = not connected. Empty list = no provider configured on this instance |
| POST | `…/{provider}/connect` | owner | `{authorize_url}` **plus a state cookie** (`HttpOnly`, `SameSite=Lax`, 10 min). Must be sent with `credentials: 'include'`, then a top-level `window.location.href = authorize_url`. 409 `already_connected` when active; allowed when `status` is `error` (reconnect) |
[...]

Issue #731, opened 2026-07-30
ci: PRs into integration branches run no checks and report as CLEAN


biagiodistefano:
## Problem

**A pull request into an integration branch runs no CI, and GitHub reports it as green.**

`.github/workflows/ci.yml`:

```yaml
on:
  push:
    branches: [main, develop, feature/*]
  pull_request:
    branches: [main, develop]
```

`pull_request.branches` filters on the **base** branch, so a PR targeting anything other than
`main`/`develop` never triggers Lint & Type Check, Translation Validation, Unit Tests or Build.
The `push` trigger is the only fallback, and it only fires for heads matching `feature/*`.

Observed live on the membership-tier work (#725–#728, all based on
`integration/membership-tier-unlock`):

| PR | head branch | CI |
|---|---|---|
| #727 | `feature/724-free-lifetime-plans` | full — passed |
| #728 | `feature/720-public-tier-grid` | full — **caught a real Translation Validation failure** |
| #725 | `fix/721-questionnaire-card-type-aware` | **none** |
| #726 | `fix/722-questionnaire-cache-invalidation` | **none** |

Two of the four were verified purely by accident of branch naming. The other two merged with
~400 lines of new, never-executed test code.

The dangerous part is the reporting: `gh pr checks` on #725/#726 showed only License/audit/Snyk,
and `mergeStateStatus` was **`CLEAN`**. "Clean" means nothing failed — not that anything ran.
There is no signal distinguishing "passed" from "never attempted".

This is not hypothetical for this repo: the integration-branch pattern is established practice
here (`feature/venue-seating`, and now `integration/membership-tier-unlock`).

Note the branch is also unchecked *after* merge — `integration/*` matches neither trigger — so
breakage only surfaces on the eventual PR to `main`, with every branch's changes tangled together.

## Proposal

Cover integration branches on both triggers:

```yaml
on:
  push:
    branches: [main, develop, 'feature/*', 'integration/*']
  pull_request:
    branches: [main, develop, 'integration/*']
```

Consider dropping the `push` branch filter for PR-bearing branches entirely and relying on
`pull_request`, so branch *naming* stops being load-bearing for whether code is tested. Check the
duplicate-run behaviour before doing so (push + pull_request both firing on the same commit).

## Acceptance criteria

- [ ] A PR based on an `integration/*` branch runs Lint & Type Check, Translation Validation,
      Unit Tests and Build.
[...]

PR #896, opened 2026-09-07
feat(integrations): phase 1 — connect an Eventbrite account (#894)

biagiodistefano:
## Summary

Phase 1 of #894: the owner-only **Integrations** page at `/org/{slug}/admin/integrations`, in the admin nav and the dashboard quick actions.

- One card per provider the instance has enabled (from the API, never hardcoded); empty state with a self-hosting docs link when there is none.
- States: not connected → **Connect**; pending → account picker (`/accounts` + `select-account`); active → "Connected as …", auto-sync checkbox, **Disconnect** (confirm dialog with the "remove Revel under Apps" note); access lost → **Reconnect**; a warning strip when live sales updates could not be enabled.
- OAuth landing: `?connected`, `?select`, `?error=<code>` rendered as strips, then stripped from the URL.
- `src/lib/utils/integration-errors.ts`: copy for every `IntegrationErrorCode`, typed as a full `Record` so a backend rename fails `make types`; falls back to the backend's translated `detail`, then a generic line. The platform's own message renders collapsed, never as the headline.
- Six locales for every new key.

## Tests

- Unit: error map (7), card state + landing helpers (10), `AccountPicker` (4), `ProviderCard` (8), tint cycle bumped to 15 owner tiles.
- `make check` and `make test` pass locally (3409 tests). PRs into the integration branch run no CI (#731).

## Not in this PR

"Import from Eventbrite" (phase 2) and the event Listings tab (phase 2).

Refs #894 · backend: letsrevel/revel-backend#924


biagiodistefano:
Verified in a browser against the backend branch (seeded E2E stack, fake Eventbrite credentials so the provider is enabled):

- Owner: page renders under the admin layout, "Integrations" in the nav and dashboard quick actions; not-connected card with **Connect Eventbrite**.
- Connect click: POST `/connect` succeeded and the browser navigated top-level to `eventbrite.com/oauth/authorize` with the signed state (Eventbrite answers 403 for a localhost redirect, as the backend spike documented).
- Landing: `?error=state_invalid`, `?error=provider_rejected` and `?connected=eventbrite` each render their strip and the param is stripped from the URL.
- Active (seeded row): "Connected as …", "since {date}", auto-sync checkbox writes `auto_sync=true` to the database and stays checked; the webhook warning strip shows for `webhook_registration_failed`.
- Disconnect: confirm dialog copy, then DELETE removes the row and the card returns to not-connected without a reload.
- Access lost (seeded `status=error`): danger card with **Reconnect** and **Disconnect**.
- Staff: the URL returns the 403 page and the nav has no Integrations item.

Two fixes came out of it and are in the branch: the not-connected card no longer repeats the platform name as its heading, and the landing param is stripped with the raw history call (SvelteKit's `replaceState` throws before the router is initialised, from `onMount` and `afterNavigate` alike).

PR #906, opened 2026-09-08
feat(attribution): carry campaign UTMs to checkout and show sales-by-source to organizers

### LINKED ISSUE #880: feat(embed): carry campaign UTMs from host page through embed to checkout, and show sales-by-source to organizers
## Context

An organizer we met (2026-09-04) markets to his own website, where Revel events are embedded via iframe, and wants to know **which of his channels sold each ticket**. Other providers lose the trail because checkout runs inside the iframe. Revel's embed never runs checkout: every CTA opens a new tab on letsrevel.io, and `src/lib/embed/links.ts` already stamps `utm_source=embed`, `utm_medium`, `utm_campaign` (org slug) and `utm_content` (host hostname, injected by `static/embed-v1.js:71`). The problem is the chain is **write-only**: nothing ever consumes those params, so an embed-sourced sale is indistinguishable from a direct one.

**Design principle (decided 2026-09-08): absolute bare minimum, zero storage on the visitor's device.** No cookie, no sessionStorage/localStorage, no `document.referrer`. Campaign tags travel in the URL only; the checkout mutations read them off the current page URL and send them to the backend, which stamps them on the ticket. Nothing else changes about tracking (there is none).

Backend counterpart (field, payloads, breakdown endpoint): letsrevel/revel-backend#922

### 1. Loader passthrough (`static/embed-v1.js`)
The loader runs on the organizer's page, so it can see that page's URL. Copy `utm_source`, `utm_medium`, `utm_campaign`, `utm_content` from `window.location.search` onto the iframe `src`. Keep the current defaults (`utm_source=embed`, `utm_content=<hostname>`) **only when the host page carries no `utm_source`** — the organizer's own campaign tags win. Read nothing else from the host page.

The embed routes then forward those params into the outbound links built by `links.ts` (today only `utm_content` is read, via `sanitizeUtmContent` in `src/lib/embed/params.ts`; extend the same allow-list sanitiser to all four keys). Embed responses stay cookie-free and shared-cacheable — the params are part of the URL, so caching is unaffected.

Document in `docs/embeds.md`: the raw-iframe snippet and the oEmbed path cannot see the host page URL, so they attribute at embed level only (`utm_source=embed` + hostname).

The loader filename is the version (immutable cache): the change only adds query params, so staying on `embed-v1.js` is probably fine — confirm in the PR.

### 2. Read at checkout — from the URL, nothing stored
In each checkout mutation, read `utm_*` from `page.url.searchParams` at the moment of the reserve call, sanitise with the shared allow-list, and pass as `attribution` (once the API client is regenerated):
- `src/lib/components/events/cart-checkout-controller.svelte.ts`
- `src/lib/components/events/guest-cart-checkout-controller.svelte.ts`
- `src/lib/components/series-passes/SeriesPassPurchaseDialog.svelte`
- (resume path in `event-checkout-controller.svelte.ts` only if the BE accepts it there; the ticket already exists at that point, so probably not)

A tiny pure helper `readAttributionFromUrl(url): TicketAttribution | null` in `src/lib/embed/params.ts` (or `src/lib/utils/`) keeps the four call sites one-liners and is unit-testable.

**Explicitly NOT:** no `handleTokenCapture` change, no cookie (the `?ref=` precedent is not followed here), no storage of any kind, no `document.referrer`.

Verify the tags actually survive the paths that matter (tests already pin most of this):
- `post-redirect-params.ts` strips only its own keys and preserves `utm_*` (`post-redirect-params.test.ts:97-101`) ✔
- login/registration redirects preserve the full `returnUrl` incl. query (`auth-guard.test.ts:58`) ✔
- cart dialogs / guest checkout stay on the event page (no navigation) — confirm nothing does a `goto()` that drops the query before reserve.

Accepted consequence: if the visitor navigates away and comes back later, the sale shows as `direct`.

### 3. Organizer UI
- Event admin: a "Sales by source" card/table fed by the new breakdown endpoint (source / medium / campaign / content / count, `direct` bucket for untagged sales). Empty state when everything is direct.
- Admin ticket list: filter by source and campaign.
- Attendee export already gains the columns server-side; nothing to do beyond regen.
- Embed builder (`src/lib/components/embed/EmbedBuilder.svelte`): one sentence telling organizers that campaign tags on their page flow through to the sales report when they use the script snippet.

## Out of scope (decided)
- Cookies, session/local storage, referrer capture, cross-session attribution.
- Any analytics vendor, page views, funnels.
- Conversion pings back to the organizer's own GA (purchase completes in another tab on our origin; not possible).

## Testing
- Unit: loader passthrough (host UTMs win, defaults otherwise, nothing but the four keys copied), sanitiser for all four keys, `readAttributionFromUrl` (missing → null, junk → dropped).
- E2E: land on an event page with `?utm_source=instagram&utm_campaign=sept`, buy a free ticket as guest, assert the ticket shows `instagram` in the admin breakdown.
- a11y on the new admin table (WCAG 2.1 AA).

## Order
Backend PR first (contract), then `pnpm generate:api`, then this.

biagiodistefano:
Closes #880. Backend counterpart: letsrevel/revel-backend#944 (**must merge & deploy first** — this PR's client is regenerated from its contract; the series-pass checkout body uses the new wrapper).

## What

First-party purchase attribution, zero storage on the visitor's device — the four `utm_*` tags travel in the URL only (no cookie, no local/sessionStorage, no referrer).

- **Embed loader passthrough** (`static/embed-v1.js`, stays v1 — additive): if the host page carries `utm_source`, its four `utm_*` params are forwarded onto the iframe verbatim and the embed's outbound links use them **replacing the whole embed convention** (`embed`/`<surface>`/`<org slug>`/`<hostname>`); otherwise behavior is byte-identical to today. Sanitisation happens server-side (`parseUtmOverride`), never in the loader.
- **Shared sanitiser** (`src/lib/utils/attribution.ts`) mirroring the backend exactly (trim → 100-cap → `[A-Za-z0-9._:-]` allow-list, drop-never-400): `sanitizeUtmValue`, `readAttributionFromUrl`, `readAttributionFromCurrentUrl`, `withUtmParams`. `sanitizeUtmContent` now delegates to it.
- **UTM carry on public surfaces**: card default-hrefs (EventCard, EventSeriesCard, OrganizationCard, EventSeriesLinkCard, OrganizationInfo) and every query-string-rebuilding navigation (/events + /organizations filters/pagination, series page sort/pagination, org-profile browse links) re-append the current URL's tags — a tagged landing two clicks from checkout still attributes. No-op when the URL carries no tags.
- **Checkout capture**: the three reserve calls (authed cart, guest cart, series pass) send `attribution` read off the page URL at reserve time; not part of the resume fingerprint; series pass adopts the new wrapper body (it previously sent no body, so this is additive).
- **Organizer UI**: "Sales by source" card on the tickets admin page fed by `GET /event-admin/{id}/tickets/attribution` — one row per tag bucket with counts, Direct bucket, click-to-filter rows applying `?utm_source&utm_campaign` to the list (server-side exact match), filtered-by chip with clear. Degrades to no card if the endpoint fails. Buckets with neither source nor campaign render as non-links (nothing to filter by).
- **A11y** (WCAG 2.1 AA): real table semantics + sr-only caption, `aria-current` + ≥3:1 left-border cue on the active row, `role="status"` live region announcing result counts on filter changes (covers the page's pre-existing filters too, CLDR-pluralised in all six locales), new token pairs added to `scripts/audit-brand-themes.py` (0 failures).
- **Docs**: `docs/embeds.md` gains an Attribution section; embed builder's script tab gains a one-line hint. i18n complete in en/de/es/fr/it/pt.

## Tests

- Unit: 3501 passing — sanitiser matrix, URL reader, link-carry helper, `parseUtmOverride`, `buildEmbedLink` override (byte-identical fallback pinned), the real `embed-v1.js` evaluated in jsdom, guest-controller payload assertions, 8 component tests on the card.
- E2E (new `j10` + `j06` specs, 4/4 green on chromium + mobile-chrome, 0 skips): seeded breakdown asserts exact bucket counts (newsletter×2, instagram×1, revel-embed×1, Direct×1) + click-to-filter round trip; tagged journey pins the carry across two client-side hops, a free purchase, and the new row appearing in the admin breakdown.
- `make check` clean (armed type-gate canary), brand-theme audit 0 failures.

## Follow-ups (deliberately deferred)

- Loader forwards any non-empty `utm_source` before server sanitisation; a junk value costs the `utm_content=<hostname>` default (spec-consistent "dumb loader" choice).
- Series-pass dialog attribution body has no dedicated unit test (type-checked + covered transitively).
- Focus-ring idiom + scrollable-table `tabindex` consistency sweep (shared with pre-existing TicketTable).

## Summary by CodeRabbit

- **New Features**
  - Added organization-wide “Sales by source” reporting in Financials, with date and event filters.
  - Added UTM attribution throughout event, organization, series, embed, and checkout journeys.
  - Embedded pages now honor valid host-page campaign tags.
  - Ticket lists display attribution details for each purchase.
  - Sales-by-source sections are collapsible and automatically open when filtered.
- **Documentation**
  - Documented embed attribution behavior and where organizers can view results.
- **Localization**
  - Added translations for reporting and attribution features in English, German, Spanish, French, Italian, and Portuguese.


biagiodistefano:
Follow-up wave per the backend note (org-wide endpoint landed in BE #944):

- **Org-wide "Sales by source"** on `/org/[slug]/admin/tickets`: all-time table by default, a since-preset control (All time / 7 / 30 / 90 days / This year → ISO `since`) and an event multi-select (→ repeated `event_ids`), both URL-param driven with server-side refetch; card reused via a new `filterable={false}` variant (plain rows — no click-to-filter target org-wide). Empty filtered results keep the controls and show a muted notice; sr-only live region announces row counts.
- **Admin ticket rows** now show a muted `source · campaign` line when the ticket carries attribution (no new column).
- Client regenerated for the org endpoint + per-row `attribution`.
- Tests: +43 unit (filterable/heading variants, preset↔ISO mapping incl. the Custom fallback, event_ids toggling, row line), new e2e spec for the org view (deterministic 2/1/1/1 under the event filter; 6/6 across both projects with the existing attribution specs), full suite 3542 green, `make check` clean.

biagiodistefano:
UI rework after live smoke testing: both breakdown instances are now collapsed-by-default disclosures (native `<details>`; the event card auto-opens when a utm filter is active), and the org-wide breakdown moved from `/admin/tickets` (back to a plain event list) to its own collapsible section on `/admin/financials` — client-side query, still URL-param driven (`since` + `event_ids`), fetch deferred until first expand. E2E retargeted accordingly (6/6 green both projects, incl. a collapsed-default + auto-open assertion and a negative check on the old location); 3546 unit tests + `make check` green.

PR #907, opened 2026-09-08
fix(api-errors): surface swallowed HTTP errors in TierForm + sweep of 16 mutationFns

### LINKED ISSUE #908: Sweep: mutationFns that don't throw on res.error treat API failures as success
## Problem

The generated API client resolves with `{ error: <parsed body> }` on HTTP errors instead of rejecting (`ThrowOnError = false`, no global override in `src/lib/api/client.ts`). TanStack Query only sets `mutation.error` / fires `onError` on a **throw**, so any `mutationFn` that returns or awaits an SDK call without checking `res.error` treats a 400/422/500 as a success: `onSuccess` runs (closing dialogs, committing optimistic updates, invalidating queries), the global "Action failed" toast in `+layout.svelte` never fires, and the user gets zero feedback.

PR #907 fixed exactly this in `TierForm` (a 400 on save silently closed the dialog, making the whole inline error panel unreachable). It is unlikely to be the only instance.

## Candidates

A heuristic scan (mutationFn bodies containing an SDK call but no `throw`) flagged these call sites. **Each needs verification** — the scan window is imperfect in both directions: `RenameTicketHolderDialog` initially matched but actually throws right after the window, while `BookmarkButton` is a confirmed true positive (it awaits both bookmark calls and ignores `error` entirely, so a failed toggle keeps the optimistic state with no rollback and no toast).

- [ ] `src/routes/(auth)/account/privacy/+page.svelte:66`
- [ ] `src/lib/components/tickets/RefundTicketDialog.svelte:178`
- [ ] `src/lib/components/profile/DietaryPreferencesManager.svelte:56,80,98`
- [ ] `src/lib/components/profile/DietaryRestrictionsManager.svelte:144,162`
- [ ] `src/lib/components/members/PlansList.svelte:135`
- [ ] `src/lib/components/members/SubscriptionDrawer.svelte:192`
- [ ] `src/lib/components/events/BookmarkButton.svelte:46` (confirmed)
- [ ] `src/lib/components/events/EventRSVP.svelte:113`
- [ ] `src/lib/components/events/cart-checkout-controller.svelte.ts:124`
- [ ] `src/lib/components/events/guest-cart-checkout-controller.svelte.ts:180`
- [ ] `src/lib/components/notifications/NotificationItem.svelte:52,80`
- [ ] `src/lib/components/notifications/NotificationDropdown.svelte:34`
- [ ] `src/lib/components/notifications/NotificationList.svelte:74`

## Fix pattern

Same as `TicketingStep` / PR #907: in every `mutationFn` wrapping a generated SDK call,

```ts
const res = await someSdkCall({ ... });
if (res.error) throw res.error;
return res.data;
```

The thrown body is the parsed JSON (`{ detail: ... }`), which the global mutation `onError` toasts via `extractErrorMessage`, and which local error panels can render via `extractErrorMessage` / `extractFieldErrors`.

For call sites with optimistic updates (BookmarkButton, notifications), verify the `onError` rollback path actually restores state once errors start propagating.

## Tests

Per fixed call site, follow the `TierForm.test.ts` recipe: mock `$lib/api/generated/sdk.gen` (importOriginal spread), resolve `{ data: undefined, error: { detail: '...' }, response }`, assert the failure is surfaced (toast/panel/rollback) and success-only side effects (dialog close, cache commit) do NOT run.

## Guardrail idea (optional)

Consider a lint rule or a shared `unwrap(res)` helper in `$lib/api` so new mutationFns can't silently regress; a grep-based CI check for `mutationFn` blocks lacking `throw`/`unwrap` may be simpler.

Found while fixing #907 (itself found reviewing letsrevel/revel-backend#945).

biagiodistefano:
## Problem

The generated API client resolves with `{ error }` on HTTP errors (`ThrowOnError = false`), so any `mutationFn` that returns/awaits an SDK call without checking `res.error` makes TanStack Query treat a 400/422/500 as a **success**: `onSuccess` side effects run (dialogs close, optimistic flips commit, success toasts fire) and the existing `onError` handlers — toasts, rollbacks, inline panels — are unreachable dead code.

Found while reviewing letsrevel/revel-backend#945 (un-pausing an ONLINE tier without Stripe Connect → new 400 `StripeNotConnectedError`): the TierForm path would have swallowed that error entirely.

## Fix

`if (res.error) throw res.error;` (the pattern `TicketingStep` already used) in every affected mutationFn — TierForm's three, plus a full sweep of the codebase (16 more across 9 files):

- **TierForm**: create / update / delete — the inline error panel (field errors, billing-info notice) is now reachable; a rejected save keeps the dialog open and shows the backend detail.
- **Account privacy page**: file delete — local no-feedback `onError` dropped (reset moved to `onSettled`) so the global mutation toast reports the failure.
- **DietaryPreferencesManager**: add / delete / update.
- **DietaryRestrictionsManager**: create food item / add / delete / update.
- **TelegramConnectionManager**: connect / disconnect — the inline OTP error now renders the backend detail via `extractErrorMessage`.
- **BookmarkButton**: bookmark / unbookmark — the optimistic flip now rolls back on failure.
- **NotificationItem** (mark read/unread), **NotificationDropdown** + **NotificationList** (mark all read) — optimistic state rolls back, error toasts fire.
- **InvitationRequestCard**: cancel — the error toast now carries the backend detail.

Verified fail-loud already (no change needed): EventEditor, EventWizard, event-image-mutations, questionnaire submission page, ResourceModal, PotluckSection, EventRSVP, both cart-checkout controllers, RefundTicketDialog, PlansList, SubscriptionDrawer.

## Tests

Every fix is covered by a test that resolves the mocked SDK call with an HTTP error body (`{ detail: ... }`) and asserts the failure is surfaced (toast / inline panel / rollback) and that success-only side effects do **not** run. Each was watched fail against the unfixed component before the fix. New specs: TierForm, DietaryPreferencesManager, DietaryRestrictionsManager, TelegramConnectionManager, InvitationRequestCard, privacy page; extended: BookmarkButton, NotificationItem, NotificationList, NotificationDropdown.

`make check` green; full unit suite green (309 files / 3558 passed + 1 todo).

Closes #908

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling across event, invitation, notification, profile, Telegram, and privacy actions.
  * Failed requests now show appropriate error messages or toasts instead of being treated as successful.
  * Optimistic UI updates are correctly reverted when actions fail.
  * Dialogs remain open when ticket, connection, or other submitted actions encounter errors.
  * Privacy file deletion failures are surfaced consistently, while affected files remain listed.

* **Tests**
  * Added coverage for failed API responses, error notifications, state rollback, and dialog behavior.

### DIFF
--- a/src/lib/components/events/BookmarkButton.svelte
+++ b/src/lib/components/events/BookmarkButton.svelte
@@ -44,11 +44,10 @@
 	// *before* mutationFn runs.
 	const mutation = createMutation(() => ({
 		mutationFn: async (next: boolean) => {
-			if (next) {
-				await eventpublicattendanceBookmarkEvent({ path: { event_id: eventId } });
-			} else {
-				await eventpublicattendanceUnbookmarkEvent({ path: { event_id: eventId } });
-			}
+			const res = next
+				? await eventpublicattendanceBookmarkEvent({ path: { event_id: eventId } })
+				: await eventpublicattendanceUnbookmarkEvent({ path: { event_id: eventId } });
+			if (res.error) throw res.error;
 		},
 		onMutate: (next: boolean) => {
 			pending = true;
--- a/src/lib/components/events/BookmarkButton.test.ts
+++ b/src/lib/components/events/BookmarkButton.test.ts
@@ -147,6 +147,25 @@ describe('BookmarkButton', () => {
 		expect(toast.success).not.toHaveBeenCalled();
 	});
 
+	it('reverts and shows an error toast when the API resolves with an HTTP error body', async () => {
+		// The generated client does not reject on HTTP errors — it resolves with
+		// `{ error }` — so the mutationFn must throw it for onError to fire.
+		const user = userEvent.setup();
+		vi.mocked(eventpublicattendanceBookmarkEvent).mockResolvedValueOnce({
+			data: undefined,
+			error: { detail: 'Nope' },
+			response: { ok: false, status: 400 }
+		} as never);
+		renderButton({ eventId: 'e1', isBookmarked: false });
+		const btn = screen.getByRole('button');
+
+		await user.click(btn);
+
+		await waitFor(() => expect(btn).toHaveAttribute('aria-pressed', 'false'));
+		expect(toast.error).toHaveBeenCalled();
+		expect(toast.success).not.toHaveBeenCalled();
+	});
+
 	describe('onlyWhenBookmarked (card indicator)', () => {
 		it('renders nothing when not bookmarked', () => {
 			const { container } = renderButton({
--- a/src/lib/components/events/admin/TierForm.svelte
+++ b/src/lib/components/events/admin/TierForm.svelte
@@ -30,7 +30,7 @@
 	import TierFormAvailabilitySection from './TierFormAvailabilitySection.svelte';
 	import TierFormSeatingSection from './TierFormSeatingSection.svelte';
 	import { authStore } from '$lib/stores/auth.svelte';
-	import { extractErrorMessage, extractFieldErrors } from '$lib/utils/errors';
+	import { extractErrorMessage, extractFieldErrors, markSilent } from '$lib/utils/errors';
 	import { tierFieldLabel } from './tier-field-labels';
 	import {
 		CURRENCY_SYMBOLS,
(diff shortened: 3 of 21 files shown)


biagiodistefano:
### Code review

Found 2 issues (both fixed in 17e523d7):

1. TierForm's three mutations threw the raw error body with no local `onError`, so the root layout's default `mutations.onError` fired its generic "Action failed" toast on top of the form's own inline error panel — the exact duplicate-toast case the `silent: true` convention exists for (see RefundTicketDialog: "silent: the error renders inline in the dialog; without the flag the global mutations.onError adds a duplicate generic toast"). Now thrown via `markSilent()`, body unwrapped so `extractFieldErrors` still sees the pydantic detail array.

https://github.com/letsrevel/revel-frontend/blob/17e523d7ca710fc0577e41897e668c1b833ac276/src/lib/components/events/admin/TierForm.svelte#L303-L317

2. `DietaryRestrictionsManager.createFoodItemMutation` has no local `onError` and is awaited via `mutateAsync` inside `handleAddRestriction`'s try/catch, which already toasts — the global default would add a second generic toast. Same `markSilent()` fix.

https://github.com/letsrevel/revel-frontend/blob/17e523d7ca710fc0577e41897e668c1b833ac276/src/lib/components/profile/DietaryRestrictionsManager.svelte#L94-L105

A reviewer also suggested all 15 other throw sites need `silent: true`; that was verified false — a probe against @tanstack/query confirmed a mutation-local `onError` replaces `defaultOptions.mutations.onError` entirely, and every other site has one. Checked for bugs, CLAUDE.md compliance, git-history context, prior-PR guidance, and code-comment invariants.

PR #920, opened 2026-09-11
fix: seat-picker → cart sync loss, and fail-fast Stripe forwarder detection

### LINKED ISSUE #918: Seat-picker → cart sync is intermittently lossy (~1 in 3, both viewports)
## Summary

After the seat picker's **Done**, the cart-side UI intermittently never materializes. The seat pick does not reach the cart group at all — Playwright reports `element(s) not found`, not a wrong or stale value.

Reproduces at roughly **1 in 3**, on **both** desktop (Desktop Chrome) and mobile (Pixel 5), and is **independent of suite load** — reproduced in isolation with `--workers=1 --retries=0`.

## Evidence

Two full E2E runs on 2026-09-11, against FE `3bb65ff0` + BE `0150f091` (v2.12.1).

| Spec | Assertion that fails | Observed |
| --- | --- | --- |
| `j06-tickets/seat-selection.spec.ts:73` | tier card shows `1 seat · edit` | hard-failed both attempts (mobile), run 2 |
| `j06-tickets/seat-selection.spec.ts:409` | `getByTestId('cart-summary-bar')` contains `EUR 20.00` | hard-failed both attempts (mobile), run 1 |
| `j06-tickets/venue-overview.spec.ts:544` | pick survives browse-and-close, then Buy | flaky, **both** viewports |
| `j07-guest-attendee/guest-seated.spec.ts:389` | mixed cart (seated + GA) | flaky, **both** viewports |

Isolated measurement of `:409` — `--retries=0 --repeat-each=3 --workers=1` → **1 failed / 2 passed**.

**Which spec draws the hard failure rotates between runs; the subsystem is the constant.** Run 1 failed `:409` and passed `:73`; run 2 did the reverse. So this is one implementation defect, not four spec defects — please resist fixing it spec-by-spec at the test layer.

## Why this matters beyond CI

A 1-in-3 loss of a seat selection at the moment the buyer commits is a revenue-path defect, not just a flaky test. The E2E flakiness is the symptom that surfaced it.

## Notes for whoever picks this up

- Entry points: `src/lib/components/tickets/CartSummaryBar.svelte` and the seat-picker / cart-group state modules feeding it.
- Several of the affected specs already wrap seat clicks in `expect(...).toPass()` retry loops — i.e. this was noticed before and papered over at the test layer. Check whether those loops are masking the same root cause.
- Likely shapes worth ruling in/out: `$effect`/`$derived` ordering where the cart write lands before a subscriber exists; an async seat-hold response resolving after the picker closes; a cart group keyed on a briefly-`undefined` value; reactivity lost by mutating a non-`$state` object.

## Baseline for verification

With `stripe listen` running, the full suite is **620 passed / 1 failed / 4 flaky / 3 skipped** in 12.2m. The fix should take that to **620+/0 failed/0 flaky/3 skipped**, confirmed over repeated runs of the four specs above (`--repeat-each` with `--retries=0`), not a single green run.

### LINKED ISSUE #919: E2E: Stripe specs time out for 240s instead of failing fast when no `stripe listen` forwarder is running
## Summary

Without a `stripe listen` webhook forwarder, the 11 Stripe-dependent E2E specs fail by **timing out** — 240s per attempt, twice, on both viewports. They produce 22 failures whose error messages point at UI locators (`"Welcome, member!"` never appears, a refund affordance never renders) and say nothing about the actual cause.

Cost measured on 2026-09-11: the same suite ran in **36.5m without** the forwarder and **12.2m with** it. Over half the wall-clock was spent timing out on a missing local dependency.

## Current behaviour

`tests/e2e/support/stripe.ts` *documents* the forwarder requirement in a comment but enforces nothing. Only the `GET /api/version` backend probe self-skips (`tests/e2e/support/skip.ts`); there is no equivalent for Stripe.

Affected: `j06` self-cancel, `j10` organizer-refunds ×2, `j22` buyer-invoices + hybrid-flow, `j23` manage-subscription ×3 + subscribe-online ×2, `j26` pass-online-cancel — 11 specs × 2 viewports.

## Proposed fix: fail fast and loudly — do NOT self-skip

A precondition check that runs once and aborts these specs immediately with a message naming the real cause, e.g.:

> Stripe webhook forwarder not detected. Run `make run-stripe` in `revel-backend` (`stripe listen --forward-to localhost:8000/api/stripe/webhook`) before this suite.

**Self-skipping would be the wrong fix here**, even though it is the obvious one. This repo has already been burned by exactly that: journey specs self-skip when the backend probe fails, and `support/skip.ts` caches the probe per worker — so one timeout silently disarmed a whole worker and Playwright reported `exit 0, 400 passed, 99 skipped` as success. A Stripe self-skip would reintroduce that failure mode on the payment paths, which are the ones we can least afford to lose coverage on silently. Failing fast keeps the coverage honest while still turning a 240s mystery timeout into an instant, actionable message.

If a skip is wanted anyway for local convenience, it must be **opt-in via an explicit env var** (so CI cannot acquire it by accident) and must be counted against the documented skip baseline of 3.

## Acceptance

- With no forwarder running, the 11 specs abort in seconds with the message above; suite wall-clock returns to roughly the 12m figure.
- With the forwarder running, all 22 tests pass exactly as they do today (verified 2026-09-11: 22/22 in 2.8m).
- The documented skip baseline stays at 3.

biagiodistefano:
Two fixes that came out of driving a full E2E pass against FE `3bb65ff0` + BE `0150f091` (v2.12.1). One is a real revenue-path bug that had been hiding as test flakiness; the other is why the suite took three times longer than it needed to.

## 1 — Seat-picker → cart sync was lossy (#918)

**This is not a flaky test. A buyer loses their seat selection roughly 1 in 3 times, at the moment they commit.**

`SeatPickerDialog.handleDone` creates the cart group correctly. That makes the cart non-empty, which mounts `CartSeatGroupHolds` in the **same flush**, building a fresh `SeatHoldController` that seeds from the warm TanStack cache. The post-tap availability refetch is fire-and-forget and the availability query has no `staleTime`, so that cache is often still the **pre-hold** snapshot (`my_holds: []`). `seedFromAvailability` hard-assigned, the live-sync `$effect` wrote `[]` through, and `cart.setSeatIds(tier, [])` means *delete the group*. `CartSummaryBar` is gated on `!cart.isEmpty`, so it unmounted — which is why Playwright reported *element not found* rather than a wrong value.

It could not self-heal: deleting the group unmounted the very component whose `adoptServerHolds()` would have repaired it once the refetch landed. `onDestroy` then saw zero holds and skipped the release, **leaking the seats server-side for their full TTL** — which can later surface as a `capacity` 409, a second failure mode wearing the first one's clothes.

The fix is two terms:

- `shouldSyncSeatIds` gains a `holdCount` term so the live-sync effect can never delete a group, and (after review) an `unresolvedClaims` term so it can never shrink one either — `cart.setSeatIds` **replaces** `seatIds`, so a seed that resolves only some of a group's claimed seats would otherwise write the shortened list straight through. Verified the premise by grepping every consumer: the only writers that legitimately empty a `user_choice` group are the picker's `handleDone` / `seatPickerCloseAction` and `CartSeatHolds`' expiry sweep, both of which call `setSeatIds` explicitly and neither of which routes through this effect.
- `seedFromAvailability` **unions** known-held ids instead of overwriting (known ids first, so `max` truncation cannot split controller and group apart). This also closes the leak — the controller now owns the seat, so `onDestroy` releases it.

**Why union rather than deferring `seeded` until `dataUpdatedAt` moves:** the deferred option makes a local invariant depend on a network event that may never arrive. An *erroring* refetch leaves `dataUpdatedAt` unmoved, `seeded` stays false forever, the live-sync effect never runs, and a transient desync becomes a permanent one. It would also couple the gate to TanStack's refetch-on-mount policy, which a future global `staleTime` default could disarm invisibly — the #704 failure shape.

### Evidence, both directions

`CartSeatGroupHolds.test.ts` renders the real component under `QueryClientTestWrapper` with the cache pre-seeded to `my_holds: []` and a group carrying `seatIds: ['seat-1']`.

- **On the unfixed code it fails with the literal reported symptom**: `expected undefined to deeply equal [ 'seat-1' ]`
- **With the fix**: 4 passed, stable across 5 consecutive runs.

Each half was then neutralised separately to prove neither term is redundant — with only the gate, the group survives but `myHolds` is wrong; with only the union, a fourth test (the group's seat absent from the warm chart entry, so the union is filtered by `validSeatIds`) is what catches it.

## 2 — Stripe specs now fail fast instead of timing out (#919)

Without `stripe listen`, 11 specs × 2 viewports failed by **timing out** — 240s per attempt, twice — behind messages that pointed at UI locators and never named the cause. Measured cost: **36.5m without the forwarder vs 12.2m with it**.

The verdict is **observed, never guessed**. `stripe listen` opens an outbound websocket and binds no local port, so nothing about it is reachable from the test process, and a `pgrep` probe would be platform-specific *and* a liar — a wedged or wrong-target forwarder still shows up. The only honest signal is the forwarder's **effect**: a backend state change no other route can produce. A false "delivered" is therefore impossible by construction; it is only ever written after a real webhook-driven assertion has passed.

**Deliberately not a skip.** This repo has already been burned by a silent self-skip reporting `exit 0, 400 passed, 99 skipped` as success. Reintroducing that on the payment paths would trade a loud timeout for silently lost coverage. Every path stays a failure; the documented skip baseline stays at **3**.

The verdict lives in a run-keyed file rather than a module variable because Playwright spawns a **fresh worker process after every test failure** — so the `support/skip.ts` per-worker cache pattern would be discarded at exactly the moment it matters, and all 22 tests would pay full price again.

Polling semantics and timeout budgets are unchanged, so there is no new flakiness risk while webhooks flow.

### Known trade-off, stated rather than buried

If the *first* webhook wait of a run expires from a genuine app regression rather than a missing forwarder, the remaining Stripe tests abort under the forwarder message. The run is red either way, and the message names the discovering test and hedges explicitly rather than asserting a diagnosis — the original assertion error is inlined and attached as `cause`. Worth knowing when reading a red run.

## Review response (second round)

A code review of this branch found a real gap in the first fix, plus two findings from CodeRabbit. All are now addressed in `0807751c` and `1cb10fd5`:

1. **The gate stopped group deletion but not group shrinking** — the same bug class as #918, one level down, and untested because all four original tests used a single-seat group. Fixed with the `unresolvedClaims` term; new regression test uses a two-seat group with a one-seat warm chart and was proven to fail before and pass after (`expected [ 'seat-1' ] to deeply equal [ 'seat-1', 'seat-2' ]`).
2. **The missing-forwarder verdict rested on an arbitrary UI assertion** — a UI regression could rule out a run under a forwarder diagnosis it never established. The first uncorroborated expiry now rethrows the caller's own assertion error verbatim and leaves the forwarder as an annotation; only two *distinct* expired effects rule out a run.
3. **The first expiry's diagnostic could be overwritten** by a parallel worker — markers are now created exclusively.

Two further review findings were investigated and **deliberately not acted on**, because they were factually wrong about the control flow: that `seedFromAvailability` re-runs on every effect pass (it is guarded by `!seeded`, which is set immediately, so it runs exactly once), and that the expiry sweep reads `controller.holdExpiresAt` (it reads `registry.expiresAt`, derived from `availabilityQuery.data.my_holds_expire_at`).

Note on CodeRabbit's suggested remedy for (1): preserving the unresolved ids through the seed would relocate the bug rather than fix it — `#reconcileHolds`, which every seat tap goes through, filters through the same `#validSeatIds`, so the buyer's next tap would drop the seat again. Its alternative suggestion, blocking the partial write, is what landed.

## Verification

- `make check` — 0 errors, type gate armed, i18n clean, all files within line limits.
- `make test` — 307 files, 3589 passed, 1 todo.
- E2E verification against the live stack is running now; I'll post results as a comment.

## Note for review

`seat-hold-controller.svelte.ts` landed at **497/500** lines. The next edit there will need the same care or it will trip the file-length gate.

Closes #918
Closes #919

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented selected seats from disappearing when reopening or synchronizing the seat picker.
  - Preserved cart groups during seat selection, including when availability data is delayed or incomplete.
  - Improved handling of previously held seats and expired groups without adding unintended seats.

- **Tests**
  - Expanded coverage for seat-hold persistence and cart synchronization.
  - Improved checkout verification and diagnostics for Stripe-powered flows.

- **Documentation**
  - Clarified Stripe webhook setup requirements and failure behavior for end-to-end testing.

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -30,6 +30,8 @@ const REGRESSION_BROWSERS = [
 
 export default defineConfig({
 	testDir: './tests/e2e',
+	// Plants E2E_RUN_ID for the cross-worker Stripe webhook verdict (#919).
+	globalSetup: './tests/e2e/support/global-setup.ts',
 	fullyParallel: true,
 	forbidOnly: !!process.env.CI,
 	// One local retry. The client-auth rotation race that originally motivated
--- a/src/lib/components/tickets/CartSeatGroupHolds.svelte
+++ b/src/lib/components/tickets/CartSeatGroupHolds.svelte
@@ -44,6 +44,15 @@
 	const tier = group.tier;
 	const isUserChoice = tier.seat_assignment_mode === 'user_choice';
 
+	// The seats this group ALREADY claims at mount (#918) — the picker's Done
+	// wrote them in the same flush that mounted this component. Captured once,
+	// deliberately, for the same reason `tier` is: it's the hand-off payload,
+	// not a value to track. Copied so a later `setSeatIds` can't retroactively
+	// change what the seed was told; fed to `seedFromAvailability` so a warm
+	// but PRE-hold availability snapshot can't make the controller disown them.
+	// svelte-ignore state_referenced_locally
+	const initialSeatIds: string[] = isUserChoice ? [...group.seatIds] : [];
+
 	// Declared ahead of `options` (explicit type annotation) so `getQuantity`
 	// can close over it without a circular type-inference error — it's only
 	// ever CALLED after the assignment below, never read synchronously.
@@ -136,23 +145,24 @@
 			// current holds keeps each group's adoption in its own lane.
 			const validIds = validSeatIdsFor(chart);
 			for (const id of registry.otherHolds(tier.id)) validIds.delete(id);
-			controller.seedFromAvailability(validIds);
+			controller.seedFromAvailability(validIds, initialSeatIds);
 			return;
 		}
 		controller.adoptServerHolds();
 	});
 
-	// Live sync: the controller's held seats ARE the group's seatIds. The
-	// ONLY other writer is the host's expiry sweep (which clears to []).
-	// Gated on `seeded` (shouldSyncSeatIds): on the very first reactive
-	// flush `controller.myHolds` is still `[]` (chart/availability are still
-	// in flight above), and `cart.setSeatIds(tier, [])` REMOVES the group —
-	// syncing before the seed/adopt effect's first pass would strip a group
-	// whose seats already exist server-side (the venue-overview hand-off)
-	// and unmount this very component before it gets the chance to adopt
-	// them. See cart-seat-sync.ts for the (unit-tested) gate itself.
+	// Live sync: the controller's held seats ARE the group's seatIds — but
+	// only ADDITIVELY. `cart.setSeatIds(tier, [])` REMOVES the group, and this
+	// effect must never be the writer that does so: before the first seed
+	// `myHolds` is still `[]` (chart/availability in flight), and after it the
+	// seed may itself have produced `[]` from a warm-but-pre-hold availability
+	// snapshot (#918) — either way, deleting here unmounts this very component
+	// and loses the group for good (no `#each` entry left to remount from).
+	// Emptying a group is owned by the two writers that do it explicitly: the
+	// seat picker and the host's expiry sweep. See cart-seat-sync.ts for the
+	// (unit-tested) gate itself.
 	$effect(() => {
-		if (!shouldSyncSeatIds(isUserChoice, seeded)) return;
+		if (!shouldSyncSeatIds(isUserChoice, seeded, controller.myHolds.length)) return;
 		cart.setSeatIds(tier, controller.myHolds);
 	});
 </script>
--- a/src/lib/components/tickets/CartSeatGroupHolds.svelte
+++ b/src/lib/components/tickets/CartSeatGroupHolds.svelte
@@ -44,12 +44,17 @@
 	const tier = group.tier;
 	const isUserChoice = tier.seat_assignment_mode === 'user_choice';
 
-	// The seats this group ALREADY claims at mount (#918) — the picker's Done
-	// wrote them in the same flush that mounted this component. Captured once,
-	// deliberately, for the same reason `tier` is: it's the hand-off payload,
-	// not a value to track. Copied so a later `setSeatIds` can't retroactively
-	// change what the seed was told; fed to `seedFromAvailability` so a warm
-	// but PRE-hold availability snapshot can't make the controller disown them.
+	// The seats this group ALREADY claims at mount (#918), written in the same
+	// flush by whichever writer created the group: the picker's Done
+	// (`SeatPickerDialog.handleDone`) or the picker-free venue-overview
+	// hand-off (the event page's `handleSelectTier`, which adopts seats the
+	// overview map already holds). Captured once, deliberately, for the same
+	// reason `tier` is: it's the hand-off payload, not a value to track.
+	// Copied so a later `setSeatIds` can't retroactively change what the seed
+	// was told. Fed to `seedFromAvailability` so a warm but PRE-hold
+	// availability snapshot can't make the controller disown them, and
+	// compared against `myHolds` below so a PARTIAL resolution can't make the
+	// live sync write a shrunken list.
 	// svelte-ignore state_referenced_locally
 	const initialSeatIds: string[] = isUserChoice ? [...group.seatIds] : [];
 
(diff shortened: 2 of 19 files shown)


biagiodistefano:
## E2E verification against the live stack

All three runs on this branch, backend at BE `0150f091` (v2.12.1).

### 1. #918 — focused, `--repeat-each=5 --retries=0`

The three affected specs (`j06 seat-selection`, `j06 venue-overview`, `j07 guest-seated`), both viewports, retries disabled so any flake surfaces as a hard failure:

**130 passed / 0 failed (3.4m)**

Against paths that were failing ~1 in 3 before the fix, 130 consecutive passes with no retry safety net.

### 2. #919 — guard behaviour with NO forwarder running

`j23-subscriptions`, chromium, `stripe listen` deliberately not started:

| | |
| --- | --- |
| First waits (one per worker, 4 concurrent) | 2.3–2.8m each — they pay the budget, as designed |
| **Every subsequent attempt** | **18–24ms** |
| Failure messages naming the forwarder | 10 of 10 |

Retries abort instantly too. Before this change these would each have burned 240s twice.

### 3. Full suite, forwarder running

**625 passed / 0 failed / 0 flaky / 3 skipped — 628 tests, 12.3m**

Skip count is **3**, the documented baseline, so the guard did not silently acquire coverage.

Progression across the three full runs:

| Run | Result | Wall-clock |
| --- | --- | --- |
| Before, no forwarder | 600 passed / 23 failed / 2 flaky / 3 skipped | 36.5m |
| Before, forwarder up | 620 passed / 1 failed / 4 flaky / 3 skipped | 12.2m |
| **This branch, forwarder up** | **625 passed / 0 failed / 0 flaky / 3 skipped** | **12.3m** |

CI on this PR is green across all checks.

biagiodistefano:
## Review round 2 — verification

Both review commits (`0807751c`, `1cb10fd5`) verified against the live stack, BE `0150f091` (v2.12.1).

| Check | Result |
| --- | --- |
| `make check` | 0 errors, type gate armed, line limits OK |
| `make test` | 307 files, **3592 passed**, 1 todo (+3 from the new regression cases) |
| Focused, `--repeat-each=5 --retries=0`, 3 seat specs × 2 viewports | **130 passed / 0 failed** |
| Full suite, forwarder running | **625 passed / 0 failed / 0 flaky / 3 skipped, 12.0m** |

Skip count still 3. Both the partial-shrink guard and the reworked corroboration logic are green with no change to the happy-path timings.

### On the two findings not acted on

Recorded here so they are not re-raised later. Both were verified against the source and are factually wrong about the control flow:

- *"`seedFromAvailability` is re-invoked on every effect run, so `knownHeld` forces the seat back in forever"* — it is guarded by `if (!seeded)` with `seeded = true` set immediately and an early `return`; the else-branch calls `adoptServerHolds()`. The seed runs exactly once.
- *"an expiry sweep keyed off `holdExpiresAt` would never catch the phantom seat"* — the sweep reads `registry.expiresAt`, a `$derived` over `controller.availabilityQuery.data?.my_holds_expire_at`. `controller.holdExpiresAt` has no production reader at all.

### Known residual

Corroboration-by-two-distinct-effects is a heuristic. Two descriptions within the same flow could in principle corroborate each other under a single UI regression. Eliminating it needs a backend affordance that does not exist today: a staff- or debug-only read over `StripeWebhookEvent` (even a count since a timestamp on `/api/healthcheck`). The table is populated — `handle_event` inserts before dispatch — but no REST route reads it, and the E2E seed creates no superuser for the admin path. Out of scope for a frontend PR; worth a backend issue if we want the guard to observe the forwarder directly.

biagiodistefano:
Filed the backend affordance that would remove the corroboration heuristic: letsrevel/revel-backend#981.

The data already exists — `StripeWebhookEvent` is written by `service/stripe_webhooks.py:98` as the idempotency token, i.e. before dispatch, so it records that a *verified* webhook arrived regardless of whether handling succeeded. There is just no read path: no controller or API route touches the table, and the admin is session auth with no superuser in the E2E seed.

Once that lands, `expectWebhookEffect` can observe the forwarder directly and `CORROBORATION_THRESHOLD` goes away. Not a blocker for this PR.

PR #939, opened 2026-09-16
feat(referral): apply page, partner entry points, invite prefill on register

### LINKED ISSUE #938: Referral program: apply page, partner entry points (footer + account), invite prefill on register
Backend: letsrevel/revel-backend#987 (referral program applications, admin invites, case-insensitive codes). This issue is the frontend half. Regenerate the API client from the backend branch first (`pnpm generate:api`).

## What the backend now does

- `GET /version` → `features.referral_applications: boolean` (from SiteSettings; default off). Everything below is **hidden when false** — no links, no page, no menu entry.
- `POST /referral/apply` (anonymous, throttled 10/day/IP) body `{ email, code, note }`. `code` matches `^[A-Za-z0-9_-]{3,20}$`, `note` is required, 1–2000 chars, plain text. Responses: `202 {"status":"ok"}` on success **and** for blocked / already-enrolled emails (deliberately indistinguishable); `409 {"detail": "You already have a pending application."}`; `409 {"detail": "This referral code is already taken."}`; `422` validation; `404` when the feature is off.
- `GET /referral/invitations/{id}` → `{ email, code }` for an approved invite that hasn't signed up yet, `404` otherwise.
- Referral codes are now case-insensitive and keep the case they were typed in (`biagio`, `Biagio-2026`). `GET /referral/validate?code=` and registration accept any case.
- Enrollment happens on the backend by email match when the invitee registers (email or SSO). The invite id is **not** sent on registration; it only drives prefill.

## Scope

1. **Apply page** (public route, e.g. `/referral/apply` — final path/terminology TBD, see below). Fields: email (prefilled when logged in), desired code, note. Show the two 409 messages inline; any 202 shows a "thanks, we'll be in touch by email" state (never reveal blocked/enrolled). 404 when the flag is off.
2. **Entry points — present but not prominent:**
   - Footer, *Solutions* list: add a last item after "Kink event ticketing" (`Footer.svelte`, `footer.solutionKink`) linking to the apply page. Copy TBD: "Become a partner" / "Ambassador program" / "Referral program" — pick one and keep it consistent with the existing `referral.*` messages.
   - Account area: one quiet entry for users who are **not** referrers yet (`!user.referral_code`) — e.g. in the profile page or the account section of `UserMenu.svelte` next to where the referral section already appears for referrers. Enrolled users keep the existing referral entries; don't show both.
   - Both gated on `features.referral_applications`.
3. **Register page**: when `?referral_invite=<id>` is present, call the invitations endpoint; on 200 prefill and lock the email and show an "You've been invited to the referral program" banner; on 404 fall back to the normal form silently. Keep `?ref=` behaviour as is.
4. **`ReferralCodeInput.svelte`**: stop `toUpperCase()` / the `uppercase` class; display codes as stored (`/account/referral` too).
5. **i18n** for all new copy in every paraglide locale; WCAG AA (labels, error association, focus on the success state); mobile-first.

## E2E (seeded by `make e2e-setup` → backend `bootstrap_test_events`)

- Flag is on in the seeded DB.
- `test.referrer@example.com` / `password123`: enrolled, code `test-partner` (lowercase — assert it renders as typed), 20% share.
- Open invite: `test.invitee@example.com`, id `aaaaaaaa-0000-4000-8000-000000000002` → `/register?referral_invite=aaaaaaaa-0000-4000-8000-000000000002` prefills that email; registering with it enrolls (check-email page as usual).
- Pending: `test.applicant@example.com` (applying again → 409 pending). Blocked: `test.blocked@example.com` (apply → 202, nothing visible). Rejected: `test.rejected@example.com` (apply → 202).
- Taken code for the 409: `test-partner`.
- Suggested specs: `j21-referral/apply.spec.ts` (happy path, both 409s, flag-off hides links) and `j02-registration/referral-invite-registration.spec.ts`.

## Out of scope

Admin decisions (approve / reject / block / invite) live in the Django admin only.

biagiodistefano:
The frontend half of the referral program (backend: letsrevel/revel-backend#987). A public application page, two quiet entry points into it, and invite-driven prefill on registration — all gated on the backend's `features.referral_applications` switch.

Closes #938

### 1. `/referral/apply` — the application page

SSR route in `(public)` with a form action, not a client-side fetch. Three reasons:

- the flag has to 404 the **route**, not just hide a button;
- `POST /referral/apply` is throttled 10/day/IP, and the SSR path is the one that carries the visitor's real IP to the backend (`handleFetch` sets `X-Real-IP`/`X-Forwarded-For`);
- it works with JavaScript off.

Fields: email (prefilled from the auth store once it hydrates), desired code, note with a live character counter. Above them, a short "How it works". Signed-in users who already have a referral code get a "you're already in" panel linking to `/account/referral` instead of a form the backend would answer with a silent 202.

Outcomes:

| Backend | UI |
| --- | --- |
| `202` | Success panel replaces the form, focus moves to its heading |
| `409` "code already taken" | Inline error on the **code** field |
| `409` "pending application" | Form-level banner (`role="alert"`) |
| `429` | "too many applications from this connection" |
| `404` | The route 404s (flag flipped off mid-session) |

Blocked and already-enrolled emails get `202` from the backend and therefore the **same success panel** as everyone else. There's an E2E test asserting exactly that, because it's a security property, not a cosmetic one.

### 2. Entry points, both flag-gated

- **Footer → Solutions**, last item: "Referral Program".
- **`/account/profile`**, a quiet section near the bottom, shown only to users who are *not* referrers yet. Enrolled users keep their existing referral entries in the user menu; nobody ever sees both.

I put the account entry on the profile page rather than in `UserMenu` because the menu is a surface every user passes through constantly, and ~100% of users are non-referrers — a permanent menu item there reads as more prominent than "present but not prominent" asks for.

### 3. `?referral_invite=<id>` on `/register`

Resolved server-side. On success: a poster-purple banner naming the code, and the email input is prefilled and **`readonly`** (never `disabled` — a disabled input submits nothing, which would post an empty email and lose the invite). Anything that doesn't resolve to a live invite — typo, consumed id, flag off, backend down — degrades silently to the ordinary form. `?ref=` is untouched.

### 4. Case-preserving referral codes

Codes are now matched case-insensitively and **stored as typed**. So:

- `ReferralCodeInput` no longer `toUpperCase()`s the value or renders it in an `uppercase` class (plus `autocapitalize="none"` so iOS stops capitalising the first letter).
- The register action's normalizer was `replace(/[^\p{L}\p{N}]/gu, '').toUpperCase()`, which turned a perfectly valid `test-partner` into `TESTPARTNER` and matched nothing. It now only trims.

## Decisions worth a second opinion

**`referral_applications` is the one fail-CLOSED feature flag.** Every other flag in `DEFAULT_FEATURES` defaults to `true` so a `/version` blip never hides a working capability. This one defaults to `false`: its surfaces 404 when the flag is off, so failing open would render a footer link and an account CTA that both dead-end. Rationale is in the code.

**Telling the two 409s apart uses text matching.** `POST /referral/apply` answers both conflicts with a bare `{"detail": "<message>"}` and no machine-readable discriminator (unlike the account-deletion 409, which carries `code: "referral_forfeiture_confirmation_required"`). The detail is `gettext`-translated per `Accept-Language` and our client forwards the UI locale, so matching only English would go dead for five of six locales — `classifyReferralConflict` matches a stable fragment of each catalog translation instead, taken verbatim from the backend `.po` files, with unit tests pinning all twelve strings. Unmatched → generic copy, never a mislabel. **A `code` field on that 409 would let all of this go away** — worth a small backend follow-up.

**The apply page is `noindex,follow`.** It only exists while the SiteSettings flag is on and 404s the moment it's switched off; an indexed URL that flips to 404 is the soft-404 signal `make audit-soft-404` exists to keep out of the index. Easy to reverse if we'd rather have it indexed.

**`src/lib/seo/build.ts` was over its 500-line cap** once the new kind landed. Rather than raising the cap, the three text-only branches (legal, auth, referral-apply) now share one `plainPageSeo` helper, extracted to `src/lib/seo/plain-page.ts`. `build.ts` went 498 → 449. There's a regression test asserting `legal` still carries **no** `robots` key while `auth`/`referral-apply` carry `noindex,follow` — the exact thing the extraction could have broken.

## i18n

All new copy in six locales. Romance locales use gender-neutral rephrasing (no glyphs), German uses the existing `*innen` convention, French uses `tu` like the rest of the referral group. Validation form errors carry stable **codes**, not prose, so the form action never decides what language the user reads.

## Testing

- `make check` — green
- `make test` — green
- Unit: `src/lib/schemas/referral.test.ts` (schema issue codes, trim→parse pipeline, all twelve backend conflict strings, invite-id guard), plus the SEO regression guard in `build.test.ts`
- E2E: `tests/e2e/journeys/j21-referral/apply.spec.ts` and `tests/e2e/journeys/j02-registration/referral-invite-registration.spec.ts`

One E2E test is destructive by nature — registering from the seeded invite consumes it — so it's ordered last in its file and commented as requiring a fresh `make e2e-setup`. `reset_events` deletes `@example.com` users, so reseeding does restore it.

## Summary by CodeRabbit

* **New Features**
  * Added a referral-program application page with instructions, forms, validation, status messages, and account navigation.
  * Added referral invitation support during registration, including prefilled and locked invited email addresses.
  * Added referral-program links in the footer and user profile.
  * Added localized referral messaging in German, English, Spanish, French, Italian, and Portuguese.
  * Referral application access is controlled by a feature flag and disabled by default.

* **Bug Fixes**
  * Referral codes now preserve user-entered casing and supported characters.


biagiodistefano:
Thanks — went through all five.

**Fixed (4):**

1. **Locked invite email vs. autofill** (Major) — correct, and worse than described: skipping the sync alone would not have been enough, because a script that writes `.value` directly also wins at submit time (the browser posts the DOM value, not the state). `syncFormValues` now *restores* the invited address rather than syncing away from it, the enhanced submit re-asserts it on the `FormData`, and the locked field opts out of `autocomplete`.
2. **Profile entry gating** — correct. Gated on `authStore.user` (which starts `null`), so an enrolled referrer no longer sees the "apply" link flash for the length of the auth bootstrap.
3. **`errorNoteTooLong` off-by-one** — correct, and not only in French: **all six** locales said "fewer than 2000" for a limit that allows 2000. Reworded to an inclusive maximum everywhere.
4. **Explicit return type on `defaultOgImageMeta`** — done, as `DefaultOgImageMeta = Pick<SeoConfig['og'], …>` so it stays tied to the shape it feeds.

**Not changed (1):**

5. **Referral throttle in E2E** — the reasoning is right for a throttled backend, but the limit is not live in E2E: `revel-backend/.env` sets `DISABLE_THROTTLING=True`, which `common.throttling.DisableableThrottleMixin.allow_request` short-circuits on, and `ReferralApplicationThrottle` inherits it. I added a comment to the spec recording that, since it is a fair thing to wonder about.

Verified after the changes: `make check` clean, `make test` 3720 passed, and the full `make e2e` suite 643 passed / 0 failed / 5 skipped (3 pre-existing + the 2 deliberate single-project pins on the seeded invite; the 2 flakies are unrelated j08/j23 local-backend transients that pass on retry).

